PR #14, opened 2022-08-31
Move from pnpm to npm

### LINKED ISSUE #8: Move away from pnpm for build of the project
`pnpm` i snot used in any other projects that the maintainers are using.
Either use `npm` used for all other VS code extensions Red Hat Fuse Tooling team maintain or `yarn` used by other Kaoto projects

apupier:
fixes #8

Next step will be to provide the package-lock.json (when the Kaoto
component will be available and that e can compile)


apupier:
in https://github.com/KaotoIO/vscode-kaoto/pull/30 the move was done to yarn

PR #15, opened 2022-08-31
1 GitHub action

apupier:
requires https://github.com/KaotoIO/vscode-kaoto/pull/14


apupier:
create PR based on branch of this organization so that jobs are exectued ont he PR, see https://github.com/KaotoIO/vscode-kaoto/pull/26

PR #16, opened 2022-08-31
Automatically start Kaoto backend

### LINKED ISSUE #2: Launch Kaoto backend automatically
Several possibilities:

- use docker image
- use jar
- use natively compiled images
- use combination of previous solutions

first iteration will be with docker image as it already exists

apupier:
- starting it using a docker image
- not handling failure
- not handling occupied port
- stopping the container when extension is deactivated
- currently starting only when editor is opening because the extension
is activated only when the editor is opening. For the future, we will
need to pre-fetch the image sooner.
- using latest image for now, using a tag will be a lot better in the
future

fixes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,7 @@ Create and edit kaoto files.
 ## Features
 
 - Create and edit Kaoto (`.kaoto.yaml` and `.kaoto.yml`) files.
+
+## Prerequisites
+
+- `docker` must be available on system path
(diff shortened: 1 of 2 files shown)


apupier:
marked as draft as i would liek to have the project compiling and build with a GitHub action before merging it

apupier:
these changes were incorporated in https://github.com/KaotoIO/vscode-kaoto/pull/30

PR #40, opened 2022-10-13
Improve message in case of error when starting the Kaoto backend with

### LINKED ISSUE #38: Provide information when docker run is not working
For instance, if docker is not installed, currently the extension is not activating with a cryptic message for end-user:
image

apupier:
Docker

slightly better but the error message in case of missing docker on system path will be very slightly better only as there will be an ENOENT error. Familiar to JavaScript/TypeScript developer but not all users. To be improved later on.

fixes #38 (we can surely do better but I think we can keep it for another iteration and another issue)

I simulated missing docker by renaming in my dev environment to "plop", so you can remove "plop" by "docker" in the image to have an idea on how it will look alike:

Screenshot from 2022-10-13 14-35-55

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -32,6 +32,7 @@ export function activate(context: vscode.ExtensionContext) {
 
   const kaotoBackendOutputChannel = vscode.window.createOutputChannel(`Kaoto backend`);
   const backendProcess = child_process.spawnSync("docker", ["run", "--rm", "-d", "-p", "8081:8081", "kaotoio/backend"]);
+  handlePotentialErrorOnKaotoBackendStart(backendProcess, kaotoBackendOutputChannel);
   const filteredOutput = backendProcess.output.filter((s) => {
       return s !== undefined && s !== null && s.length > LENGTH_OF_DOCKER_CONTAINER_ID;
     });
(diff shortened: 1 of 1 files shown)


src/extension/extension.ts
```diff
@@ -61,6 +62,28 @@ export function activate(context: vscode.ExtensionContext) {
   console.info("Extension is successfully setup.");
 }
 
+function handlePotentialErrorOnKaotoBackendStart(backendProcess: child_process.SpawnSyncReturns<Buffer>, kaotoBackendOutputChannel: vscode.OutputChannel) {
+	if (backendProcess.error !== undefined) {
```
kahboom, line 66: Not important for this PR, but for future reference you can also do this as a shorthand, and for all the other checks below: `if(backendProcess.error)` and it will  evaluate to `false` if `backendProcess.error` is: `undefined`, `null`, `NaN`, `false`, `0`, `empty`, or `""` if it's a string

kahboom (approved):

src/extension/extension.ts
kahboom: oh you didn't have to make the changes, it was just a tip, but thank you :)

PR #41, opened 2022-10-13
Providing initial test

### LINKED ISSUE #7: Provide first basic test checking that the Kaoto UI is displayed
- setup structure to play test
- basic scenario:
   - open a my.kaoto.yaml file
   - check that the UI is correctly loaded

apupier:
fixes #7

requires https://github.com/KaotoIO/vscode-kaoto/pull/32

### DIFF
--- a/it-tests/BasicFlow.test.ts
+++ b/it-tests/BasicFlow.test.ts
@@ -1,0 +1,51 @@
+import { By, CustomEditor, EditorView,  until,  VSBrowser, WebDriver, WebView } from 'vscode-extension-tester';
+import { assert } from 'chai';
+import * as path from 'path';
+
+describe('Kaoto basic development flow', function () {
+	this.timeout(120000);
+
+	const workspaceFolder = path.join(__dirname, '../test Fixture with speci@l chars');
+
+	let driver: WebDriver;
+
+	before(async function() {
+		this.timeout(20000);
+		driver = VSBrowser.instance.driver;
+		await VSBrowser.instance.openResources(workspaceFolder);
+	});
+
+	after(async function() {
+		const editorView = new EditorView();
+		await editorView.closeAllEditors();
+	});
+
+	it('Open "empty.kaoto.yaml" file and check Kaoto UI is loading', async function () {
+		this.timeout(60000);
+		await VSBrowser.instance.openResources(path.join(workspaceFolder, 'empty.kaoto.yaml'));
+		const kaotoEditor = new CustomEditor();
+		assert.isFalse(await kaotoEditor.isDirty(), 'The Kaoto editor should not be dirty when opening it.');
+		const kaotoWebview = new WebView();
+		await driver.wait(async () => {
+			try {
+				await kaotoWebview.switchToFrame();
+				return true;
+			} catch {
+				return false;
+			}
+		}, 50000); // if it is the first load, it can take a lot of time as it downloads the docker image
+		await checkPartOfTopBarLoaded(driver);
+		await checkCanvasLoaded(driver);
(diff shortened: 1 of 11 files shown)


apupier:
current failure is due to the fact that kaoto-ui is not published and vscode-extension-tester is running a `yarn list --prod --json` which is not taking care of potential `yarn link`ed dependencies (at least if there isn't an initial published version)
asked if we can find a workaround https://coreos.slack.com/archives/C047NGXPRG8/p1666342896003189

apupier:
the test itself is very very minimal. It is testing only that an element
from the top bar and one from the Canvas are loading. I think it is
enough to catch main regressions and validate a first iteration of the
test structure. of course, more advanced use cases will need to be
covered.

Technical:
- use xvfb github action. It is required for an unknown reason. Several
other VS Code test extension are not using it. Note that it is required
for Linux only.
- there is a long wait condition, it is the time needed for docker image
to be downloaded.
- remove Windows and MacOS because cannot play test on them due to
Docker issue:
    - On MacOS, Docker not available at all
    - On Windows: docker: image operating system "linux" cannot be used
on this platform.

it-tests/BasicFlow.test.ts
```diff
@@ -0,0 +1,51 @@
...
+		await VSBrowser.instance.openResources(workspaceFolder);
+	});
+
+	after(async function() {
+		const editorView = new EditorView();
+		await editorView.closeAllEditors();
+	});
+
+	it('Open "empty.kaoto.yaml" file and check Kaoto UI is loading', async function () {
+		this.timeout(60000);
```
kahboom, line 24: Does the docker image need to be downloaded for each assertion? Or why are we having so many different timeouts?

it-tests/BasicFlow.test.ts
```diff
@@ -0,0 +1,51 @@
...
+		await checkPartOfTopBarLoaded(driver);
+		await checkCanvasLoaded(driver);
+		await kaotoWebview.switchBack();
+		assert.isFalse(await kaotoEditor.isDirty(), 'The Kaoto editor should not be dirty after everything has loaded.');
+	});
+
+});
+
+async function checkCanvasLoaded(driver: WebDriver) {
+	await driver.wait(until.elementLocated(By.xpath("//div[text()='ADD A STEP']")));
```
kahboom, line 46: can we not use normal HTML selectors to check for the existence of something in the DOM, or is there a reason why we are using the driver + xpath? sorry I'm not too familiar with this approach so I may just need some enlightenment :)

it-tests/BasicFlow.test.ts
apupier: Need to be downloaded only one time.

This timeout is globally for the test execution.
We need to specify timeout for some specific driver.wait as default timeout for some is at 5s (if i remember well). The sum of all timeout in driver.wait should be lower than the global one.

it-tests/BasicFlow.test.ts
apupier: I used one way that is working. From my  far away experience several years ago), it was good habit to use either a `By.id` or a `By.xpath`. I'm open to use something else if it is better.

What is an HTML Selector?
The `By.name` https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_By.html ?

kahboom: An HTML selector, at least in this case, would be the other functions available, where you would select by the native HTML elements (e.g. by `id`, by `class`). I think if you are more comfortable with `By.xpath` then that should be fine too, it's simply that I am not used to it. :)

kahboom (approved):

Issue #70, opened 2022-11-09
Use main tag for kaoto-backend in GitHub Workflow using main kaoto-ui


apupier:
to ensure kaoto-ui and kaoto-backend are in phase.

~it requires a way to specify the docker image to use https://github.com/KaotoIO/vscode-kaoto/issues/37~

it requires to have the native executables configurable at build time or support downloading other versions.

If done at build time, I guess we can use an environment variable which will override a default version provided in the postinstall script

apupier:
obsolete with Kaoto 2 which doesn't have a backend

PR #85, opened 2022-11-15
84 download native embedded linux at build time

### LINKED ISSUE #84: Download native backend at build time
currently native binary is pushed in source code but it is a large file, would be better to download it at build time

apupier:
requires https://github.com/KaotoIO/vscode-kaoto/pull/83

fixes #84


package.json
```diff
@@ -117,6 +118,7 @@
     "css-loader": "^6.7.1",
     "file-loader": "^6.2.0",
     "mocha": "^10.1.0",
+    "node-fetch": "2",
```
apupier, line 121: node-fetch 2 is used because node-fetch 3 requires ESM support and it is not working with the way I call the script (in package;json using node) but I do not know how to launch in a different way. Open to any other library that can download a file. This one seemed used a lot and is relatively easy to use.

scripts/postinstall.js
```diff
@@ -0,0 +1,15 @@
...
+  const fileStream = fs.createWriteStream(path);
+  await new Promise((resolve, reject) => {
+      res.body.pipe(fileStream);
+      res.body.on("error", reject);
+      fileStream.on("finish", resolve);
+  });
+  fs.chmodSync(path, fs.constants.S_IRWXU);
+});
+
+downloadFile("https://github.com/KaotoIO/kaoto-backend/releases/download/v0.4.0/kaoto-linux-amd64", "./binaries/kaoto-linux-amd64");
```
kahboom, line 15: is there any way we can use an environmental variable or something to not have to hardcode the version here?

apupier: I hope yes. We will need to make it configurable in the context of this other task: #70

This is the only place where the version of kaoto-backend is defined and impact the build. So it doesn't seem needed for now. There is another reference to this version which is part of the readme, not impacting the build. I think there is no built-in support for variables in [GitHub Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/about-writing-and-formatting-on-github). Also note that the version is different than the kaoto-ui one.

kahboom (approved):
sorry for the delay, thanks @apupier !

PR #121, opened 2023-01-27
Forward stderr information to Kaoto Backend output channel

apupier:


apupier:
will investigate MacOS failure with another task

PR #123, opened 2023-01-27
Use RHEL 9 for JenkinsFile

apupier:
RHEL 8 is no more working as glibc 2.32 seems to be required by Kaoto Backend 0.6.1


apupier:
- main branch already have incompatible change it seems
- MacOS is investigated on another side. Code change is not affectign at all this part

PR #162, opened 2023-03-02
Provide how to use a different Kaoto backend during development

### LINKED ISSUE #160: Document how to use a different Kaoto backend during development
in contributing guide.
in a nutshell:
- in extension.ts file comment the part responsible to launch the backend binary and waiting for the warmup
- launch kaoto backend (on localhost:8081) the way you want
- launch the VS Code extension

Will be easier when we will be able to plug to a differen tendpoint but that's for later

apupier:
fixes #160


kahboom (approved):
perfect, thanks!

PR #163, opened 2023-03-02
Fix step extension loading

### LINKED ISSUE #157: Choice config tab is not working with Kaoto 0.7.0 `Cannot read properties of null (reading 'useState')`
image

apupier:
the trick is that the libraries were not visible to Step Extensions
because it is using micro-frontend patterns an it requires module
federation.

This commit is picking all dependencies and using them in federated
module with an `eager` state to true. This is not optimal. it can
potentially cause performance issue but none are visible currently (and
at least it is still better than crashing the editor when there is the
need for a step extension)

fixes https://github.com/KaotoIO/vscode-kaoto/issues/157

note: need to enclose the glob pattern for launching multiple tests


kahboom (approved):
tested locally and it works fine for me. thanks @apupier !

PR #170, opened 2023-03-06
Provide basic test for modify, save and reload of Camel route

### LINKED ISSUE #99: Provide test for modify and save
currently only opening and check of Kaoto UI is loaded has been provided.

The following iteration should cover modify, save and reload to check that it has been correctly saved.

### LINKED ISSUE #173: GitHub actions job on MacOS is hanging more and more often
it sounds like when the tests are failing, the job is often hanging several hours.

It sounds like when the tests are failing it is because the Kaoto backend failed to start complaining about port 8081 to be occupied. So implementing https://github.com/KaotoIO/vscode-kaoto/issues/60 could help

see also https://github.com/KaotoIO/vscode-kaoto/issues/125

apupier:
fixes #99
fixes #173 - seems was a combination of timeouts and redraw of elements that were no more clickable mainly. And the fact that it was hanging due to switchToFrame retried which was trying again and again without being "timeouted". Done in the same PR as it is imbricated and required.


apupier:
currently failing with main branch of Kaoto UI because of API break in kie-tools, we will need to have a new release of Kaoto UI and integrate it to fix that.

PR #174, opened 2023-03-08
Remove limitation to have port 8081 available

### LINKED ISSUE #60: Automatically searching for a free port for kaoto backend
currently, if the port is occupied, it won't work. it would be nice to search for a free port automatically and to start Kaoto backend with it.

for instance retrieval of free port in an old version of VS Code AtlasMap: https://github.com/jboss-fuse/vscode-atlasmap/blob/c2a7942b2795e4d7c6355470c2d49271fd2b02dc/src/extension.ts#L67

apupier:
it is delegating the choice of the port to Quarkus framework.

fixes #60

TODO:
- find a way to tell Kaoto UI to use this port
- remove hardcoded port in webpack.config.js
- provide test occupying port 8081

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ Kaoto UI is embedded in version 1.0.0. Kaoto backend is launched natively using 
 ## Limitations
 
 - Requires `Amd64` architecture
-- Port 8081 must be available
 - Kaoto files are always written and overwritten with Linux-style end of line
 - Unsupported elements by Kaoto are removed from the files when opening them. The editor will open in dirty state in this case.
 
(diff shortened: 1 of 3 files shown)


apupier:
requires https://github.com/KaotoIO/kaoto-ui/pull/1687

but currently not working. it sound like the file KaotoEditorEnveloppeApp.js is sent inside a request. I do not have debug capapbilities from the IDe then and do not know what is happening exactly. I'm suspecting that the way to access the field of our VS Code extension class needs to be done in a specific way.

apupier:
outdated, next version planned will be based on Kaoto 2 which won't require a backend and so no more port issue

PR #249, opened 2023-04-25
feat(vscode): Provide KaotoUI version

lordrip:
This is the counterpart of the [UI pull request](https://github.com/KaotoIO/kaoto-ui/pull/1692)

relates: KaotoIO/kaoto-ui#696

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,6 +5,8 @@ const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const path = require("path");
 const BG_IMAGES_DIRNAME = "bgimages";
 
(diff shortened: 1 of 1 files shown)


apupier:
works in VS Code with https://github.com/KaotoIO/kaoto-ui/pull/1692

image

little thing is that it doesn't respect the theme. I have a light theme and ti is displayed with black background (but I guess it is not specific to VS Code part?)

lordrip:
It seems that the black background it's from the about modal itself, I couldn't find a light version in [the docs](https://www.patternfly.org/v4/components/about-modal).

That being said, probably we could override the background depending on whether we're in light or dark mode.

If you think that this is mandatory, I would be happy to do it on the KaotoUI side.

apupier (approved):

PR #256, opened 2023-05-02
Change port of embedded Kaoto backend to 8097

### LINKED ISSUE #253: request: change default port to something other than 8081
Hi, can we change the default port to something a little more unique than 8081?  This port is also used by default by eg. Quarkus unit tests so every time run a Quarkus build when VSCode is running it fails unless I specifically change the port in my application.properties which isn't ideal.

apupier:
8081 port is the default one for Quarkus applications causing often port conflicts.
To mitigate this problem changed the port to 8097.

plan is to pick a free port dynamically in another iteration, see #60 

fix #253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # 0.6.0
 
+- Change default port of emebedded Kaoto backend from `8081` to `8097`
+
 # 0.5.0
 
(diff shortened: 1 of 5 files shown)


apupier:
ok, planning it for next week (today is my last workday of the week)

Awaiting for an official release, you can pick a snapshot version: https://github.com/KaotoIO/vscode-kaoto/wiki/Testing-bleeding-edge-version

tadayosi:
@djelinek @apupier Thanks!  Now I see `8097` is used. :+1:

Issue #270, opened 2023-06-07
Upgrade to Yarn v3


apupier:
it seems that we must upgrade VS Code Kaoto too due to bugs in yarn v1 and react mismatch with Kogito.

follow-up on https://github.com/KaotoIO/kaoto-ui/issues/1588

Also the steps to use locally latest version will need to be updated due to `yarn link` without arguments does not exist anymore:
https://yarnpkg.com/cli/link
https://github.com/KaotoIO/vscode-kaoto/blob/main/CONTRIBUTING.md#how-to-manually-test-latest-kaoto-ui

lordrip:
It seems that `vscode/vsce` doesn't support yarn v3 yet

 https://github.com/microsoft/vscode-vsce/issues/517

lordrip:
We now have some situations to consider:

* Scenario # 1: `kaoto-ui@1.0.0` + `kaoto-backend@1.0.0` + `@kie-tools/*@0.27.0`
Everything works fine, no problems whatsoever

* Scenario # 2: `kaoto-ui@main-branch` + `kaoto-backend@1.0.0` + `@kie-tools/*@0.27.0`
    * `vscode-kaoto` cannot link `kaoto-ui` because `yarn v3` checks whether the dependencies are correct or not, upon linking projects (_this didn't happen in the past, so weird stuff occurred from time to time_)
    * Apart from the linking issue, the latest UI requires the latest backend, otherwise, the endpoints that support multi-routes aren't available (`v2/`)

* Scenario # 3: `kaoto-ui@main-branch` + `kaoto-backend@1.0.0` + `@kie-tools/*@0.29.0`
   * `vscode-kaoto` can now link `kaoto-ui`
   * `vscode-kaoto` doesn't compile because there's a breaking change between `@kie-tools/*@0.27.0` and `@kie-tools/*@0.29.0`
   * the latest UI requires the latest backend, otherwise, the endpoints that support multi-routes aren't available (`v2/`)

With all that said, what would be the best approach in your opinion to move forward with `kaoto-ui@main-branch` tests?

lordrip:
@apupier Now that I think of it, it seems to me that usually our problems come from the dependencies mismatch, and probably a big part of that it's because of how we're interacting with `@kie-tools`.

I haven't tested it yet, but I believe that we could move everything regarding `@kie-tools` into `vscode-kaoto` and offer hooks to provide what would be missing from the interconnection point of view.

[...]

PR #273, opened 2023-06-08
chore: update yarn to v3

lordrip:
### Context
Upgrade `yarn` to the `v3` version

### Changes
* Update `yarn` version
* Update `@kie-tools-core/*` dependencies to match with `kaoto-ui` version
* With the invaluable help of @djelinek, refactored the tests to account for the delay when opening VSCode in CI.

### Known issues
* `kaoto-ui@main` is not working since it requires the latest version of `kaoto-backend` since it provides the new `/v2` endpoints for the multiple routes support.

fixes: https://github.com/KaotoIO/vscode-kaoto/issues/270


lordrip:
@apupier I have the following situation whenever we link the latest `kaoto-`ui` with `vscode-kaoto`: apart from the `@kie-tools` dependency mismatch, the latest version of `kaoto-ui` requires the latest version of the backend, which is not released yet :grimacing: 

Should we stay then in the released `kaoto-ui` when linking locally then?

apupier:
> Should we stay then in the released `kaoto-ui` when linking locally then?

I guess you mean the `kaoto-backend`.
it would be nice to use the latest kaoto-backend in jobs but it will require to build the native binaries locally as it is not published which takes quite some time and will be complicated to maintain. So for now, I let it aside and relied on frequent releases of backend and ui.
Corresponding issue https://github.com/KaotoIO/vscode-kaoto/issues/70

When developing locally, there is a little trick to use another backend: https://github.com/KaotoIO/vscode-kaoto/blob/main/CONTRIBUTING.md#how-to-manually-test-latest-kaoto-backend

lordrip:
> > Should we stay then in the released `kaoto-ui` when linking locally then?
> 
> I guess you mean the `kaoto-backend`. it would be nice to use the latest kaoto-backend in jobs but it will require to build the native binaries locally as it is not published which takes quite some time and will be complicated to maintain. So for now, I let it aside and relied on frequent releases of backend and ui. Corresponding issue #70
> 
> When developing locally, there is a little trick to use another backend: https://github.com/KaotoIO/vscode-kaoto/blob/main/CONTRIBUTING.md#how-to-manually-test-latest-kaoto-backend

I agree, it would be nice to have the latest Kaoto backend. In this particular case, using the main branch for Kaoto UI won't work with the backend 1.0 because of the former using newer `/v2` endpoints from the latter to support multiple routes.

src/webview/KaotoEditorEnvelopeApp.ts
```diff
@@ -1,10 +1,10 @@
...
+import { Editor, EditorFactory, KogitoEditorChannelApi } from '@kie-tools-core/editor/dist/api';
+import * as EditorEnvelope from '@kie-tools-core/editor/dist/envelope';
 declare const acquireVsCodeApi: any;
 
 EditorEnvelope.init({
-  container: document.getElementById("envelope-app")!,
+  container: document.getElementById('envelope-app')!,
   bus: acquireVsCodeApi(),
-  editorFactory: new KaotoEditorFactory(),
+  editorFactory: new KaotoEditorFactory() as unknown as EditorFactory<Editor, KogitoEditorChannelApi>,
```
lordrip, line 9: There's a mismatch between `@kie-tools-core` from `kaoto-ui` and `vscode-kaoto`. This will be removed during the next release of `kaoto-ui`.

package.json
```diff
@@ -159,14 +159,17 @@
     "tsconfig-paths-webpack-plugin": "^4.0.1",
     "typescript": "^5.0.4",
     "url-loader": "^4.1.1",
-    "vscode-extension-tester": "5.6.0",
+    "vscode-extension-tester": "^5.7.1",
```
lordrip, line 162: Upgraded this dependency as it has [bigger timeouts](https://github.com/redhat-developer/vscode-extension-tester/commit/9d472b09844c6000e45d064a4b38835842d5aebe#diff-f6d013860802040afba4b358dffd5b65468f48aa00377f49d00636bac2da904bR79)

[Changelog](https://github.com/redhat-developer/vscode-extension-tester/compare/v5.6.0...v5.7.1)

package.json
```diff
@@ -31,13 +31,13 @@
...
   "dependencies": {
     "@kaoto/kaoto-ui": "1.0.0",
-    "@kie-tools-core/backend": "^0.27.0",
-    "@kie-tools-core/editor": "^0.27.0",
-    "@kie-tools-core/i18n": "^0.27.0",
-    "@kie-tools-core/monaco-editor": "^0.27.0",
-    "@kie-tools-core/patternfly-base": "^0.27.0",
-    "@kie-tools-core/vscode-extension": "^0.27.0",
-    "@kie-tools-core/workspace": "^0.27.0",
+    "@kie-tools-core/backend": "^0.29.0",
```
lordrip, line 34: This upgrade is needed to match the `kaoto-ui` current dependencies, as it is a requirement from `yarn link` to match dependencies.

it-tests/Util.ts
```diff
@@ -2,31 +2,59 @@ import { assert } from 'chai';
...
+    1000
+  );
+  return { kaotoWebview, kaotoEditor };
 }
 
 export async function checkEmptyCanvasLoaded(driver: WebDriver) {
-	await driver.wait(until.elementLocated(By.xpath("//div[@data-testid='viz-step-slot']")));
+  await driver.wait(until.elementLocated(By.xpath("//div[@data-testid='viz-step-slot']")));
+}
+
```
lordrip, line 46: Moved the workaround here to ensure that we have time after each workspace loading.

There's an open feature request in redhat-developer/vscode-extension-tester#854 that could help with this.

package.json
apupier: surprised that it is working with kaoto-ui 1.0.0.

.yarn/plugins/plugin-list.js
```diff
@@ -0,0 +1,138 @@
+// Workaround for https://github.com/microsoft/vscode-vsce/issues/517
+// taken from https://gist.githubusercontent.com/arendjr/415747652a8c79f12b005ce3cfb2f808/raw/a4a227070319e1d4c61e982b74153765d6338694/list-plugin.js
+
+const fs = require('fs');
+
+// Setup: Place this file in `.yarn/plugins/list-plugin.js` and the following
```
apupier, line 6: the path indicated is not the same than the one it is really used list-plugin.js <-> plugin-list.js

src/webview/KaotoEditorEnvelopeApp.ts
apupier: So related to the comment here https://github.com/KaotoIO/vscode-kaoto/pull/273/files?diff=split&w=1#r1229554845

if I understand well, this is a trick to avoid failure at compile time but in case a method that does not exists is called it will fail at runtime?

apupier (commented):
- The jenkinsfile must be updated too given https://github.com/KaotoIO/vscode-kaoto/blob/main/Jenkinsfile

apupier:
build success with slighlty modified script to test on Jenkins: https://studio-jenkins-csb-codeready.apps.ocp-c1.prod.psi.redhat.com/job/Fuse/job/VSCode/job/vscode-kaoto-release/104/console

apupier (approved):

Issue #288, opened 2023-06-21
Right background image is missing in about dialog (main branch Kaoto UI)


apupier:
in VS Code:
![image](https://github.com/KaotoIO/vscode-kaoto/assets/1105127/e8d702d0-b678-4527-b5fd-b29b5d80f5d7)

in Kaoto snapshot:
![image](https://github.com/KaotoIO/vscode-kaoto/assets/1105127/635906f3-da28-4eb1-b915-ad3f81f3f249)

PR #291, opened 2023-06-23
fix(vscode): Parse dist/lib assets using asset/resource

lordrip:
### Context
The `url-loader` loader from `webpack` was deprecated and it's writing instead:

```
module.exports = __webpack_public_path__ + "image-filename.jpg";
```

### Changes
The fix is to switch to the asset/resource configuration which replaces the deprecated API.

### Notes
Webpack docs: https://webpack.js.org/guides/asset-modules/

fixes: https://github.com/KaotoIO/vscode-kaoto/issues/288


webpack.config.js
```diff
@@ -249,40 +249,7 @@ module.exports = async (env) => [
...
-                (input) =>
-                  posixPath(input).includes(
-                    "node_modules/@patternfly/react-inline-edit-extension/node_modules/@patternfly/react-styles/css/assets/images"
-                  ),
-              ],
-            },
-          ],
-          use: [
-            {
-              loader: "url-loader",
```
apupier, line 278: Shouldn't we remove the url-loader from devDependencies then?

lordrip: Yes, you're right, I got confused about whether `url-loader` was already part of `Webpack`. I removed the deprecated loaders from the `devDependencies` array and adjusted the configuration to use `asset`.

apupier (approved):

PR #299, opened 2023-07-11
Do not embed node_modules in vsix

### LINKED ISSUE #187: Configure webpack to not embed node_modules
Currently, node_modules folder is embedded in .vsix. it is taking a huge place. Removing it was breaking the vsix so kept it as initial iteration.

thoughts:
- what about inheriting most of the webpack config from kaoto-ui?
- check that tests will be launched with a webpacked vsix
- I guess current build:prod target was supposed to deal with that

apupier:
- not using `yarn pack:prod` as finer grained is required with JenkinsFile
- adapted GitHub Actions to be as close as possible as JenkinsFile

fixes #187

It allows a size reduction of packed vsix from 244.3Mo to 104.5Mo

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,9 +20,8 @@
   "scripts": {
     "lint": "yarn eslint ./src --ext .ts,.tsx",
     "postinstall": "node ./scripts/postinstall.js",
-    "build:prod": "rimraf dist && webpack && yarn pack:prod",
+    "build:prod": "rimraf dist && webpack",
     "build:dev": "rimraf dist && webpack --env dev",
-    "pack:prod": "vsce package --githubBranch main --no-dependencies --yarn -o ./dist/vscode_kaoto_$npm_package_version.vsix",
     "compile": "webpack",
     "watch": "webpack --env dev",
     "run:webmode": "yarn vscode-test-web --browserType=chromium --extensionDevelopmentPath=. --open-devtools ./resources",
--- a/package.json
+++ b/package.json
@@ -161,7 +161,8 @@
     "web-worker": "^1.2.0",
     "webpack": "5.78.0",
     "webpack-cli": "5.0.1",
-    "webpack-merge": "^5.8.0"
+    "webpack-merge": "^5.8.0",
+    "webpack-permissions-plugin": "^1.0.9"
   },
   "resolutions": {
     "@types/react": "18.2.13",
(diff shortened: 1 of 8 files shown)


apupier:
- tests are not played with the packed version
- the embedded binaries does not have anymore the execution right

package.json
```diff
@@ -20,13 +20,13 @@
...
     "postinstall": "node ./scripts/postinstall.js",
-    "build:prod": "rimraf dist && webpack && yarn pack:prod",
+    "build:prod": "rimraf dist && webpack",
     "build:dev": "rimraf dist && webpack --env dev",
-    "pack:prod": "vsce package --githubBranch main --no-dependencies --yarn -o ./dist/vscode_kaoto_$npm_package_version.vsix",
     "compile": "webpack",
     "watch": "webpack --env dev",
     "run:webmode": "yarn vscode-test-web --browserType=chromium --extensionDevelopmentPath=. --open-devtools ./resources",
     "test:it": "tsc --project tsconfig.it-tests.json --skipLibCheck --sourceMap true && extest setup-and-run --yarn --uninstall_extension --extensions_dir ./test-resources 'out/*.test.js'",
+    "test:it:with-prebuilt-vsix": "extest get-vscode && extest get-chromedriver && extest install-vsix --vsix_file vscode-kaoto-$npm_package_version.vsix --extensions_dir ./test-resources && extest run-tests --extensions_dir ./test-resources 'out/*.test.js'",
```
apupier, line 29: @mlorinc I created a separated script to be able to launch test with packed with production on CI

apupier: The existing script was not deleting the --extension-dir so sounds fine to keep the same behavior.

mlorinc: test:it has --uninstall_extension, so it is removed in background. Directory itself is not removed, however the extension is.

apupier:
oups, the tests are not played:

https://github.com/KaotoIO/vscode-kaoto/pull/299/checks#step:10:23

package.json
apupier: I think we cannot use --uninstall_extension otherwise, it will remove the one that we have just installed and want to test

mlorinc: It will be uninstalled in after all Mocha Hook (handled by extester).

https://github.com/redhat-developer/vscode-extension-tester/blob/672c58cf995924df748911414e12bc2a69c331e9/src/suite/runner.ts#L91

https://github.com/redhat-developer/vscode-extension-tester/blob/672c58cf995924df748911414e12bc2a69c331e9/src/util/codeUtil.ts#L211

unsortedhashsets (approved):
`It allows a size reduction of packed vsix from 244.3Mo to 104.5Mo` WOW
Very nice idea, lgtm

mlorinc (approved):

PR #319, opened 2023-07-24
Improve GH actions workflow tests stability

djelinek:
EDIT:
- without this PR the tests failures are in place almost for each run

- with these changes the port in use issue is gone and also context menu item issue ia fixed


djelinek:
It seems I was able to fix root issue with sometimes the port 8097 was used which caused initial startup of vscode instance and right after window reload caused by `openResources` extester method used for open testing workspace

but now there is a quite often failing open context menu test

PR #430, opened 2023-12-21
Re-enable check for integration name in top bar dropdown

djelinek:
- update to Kaoto 0.7.0
- bump ExTester to latest version
- small fix for BasicFlow test on L128


djelinek:
not sure why but naming of integration is not properly displayed and working with current release of kaoto-next, on Main branch it is fixed and working as expected - reproduced also manually

djelinek:
> not sure why but naming of integration is not properly displayed and working with current release of kaoto-next, on Main branch it is fixed and working as expected - reproduced also manually

fixed with upgrade to latest Kaoto next release v0.7.0

apupier (approved):

PR #431, opened 2023-12-21
Add initial tests for switching between editor tabs

djelinek:
https://issues.redhat.com/browse/FUSETOOLS2-2271

### DIFF
--- a/it-tests/SwitchBetweenTabs.test.ts
+++ b/it-tests/SwitchBetweenTabs.test.ts
@@ -1,0 +1,101 @@
+import { By, EditorView, until, VSBrowser, WebDriver, WebElement, WebView } from 'vscode-extension-tester';
+import { openAndSwitchToKaotoFrame } from './Util';
+import { join } from 'path';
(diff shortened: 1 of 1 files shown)


it-tests/SwitchBetweenTabs.test.ts
```diff
@@ -0,0 +1,76 @@
...
+			driver,
+			true
+		)).kaotoWebview;
+		expect(await getTabsCount()).to.equal(2);
+		expect(await getActiveTabName()).to.equal('Design');
+	});
+
+	it('Switch to "Beans" tab and check it is active', async function () {
+		await switchToTab(locators.EditorTabs.beansTabLabel);
+		expect(await getActiveTabName()).to.equal('Beans');
```
apupier, line 48: I think we need to check that there is some content loaded

djelinek: added wait for elements are located

apupier (approved):

Issue #455, opened 2024-01-23
Support opening `*.camel-k.yaml` and `*.pipe.yaml`


igarashitm:
The file name `*.camel-k.yaml` and `*.pipe.yaml` enables the vscode text editor to validate the syntax against Camel K CRD. We'd also allow opening it with Kaoto.
https://github.com/KaotoIO/vscode-kaoto/blob/main/package.json#L57-L65

apupier:
* it is [already possible](https://github.com/KaotoIO/vscode-kaoto/blob/4a98a9e90dbee462c69be080a7736248d2515335/package.json#L88) to open all `*.yaml` via right-click. In theory contextual menu should be provided even without activation
*` *.camel-k.yaml` and `*.pipe.yaml` are not standard naming from Camel community project. I do not think that we should list them directly.

PR #456, opened 2024-01-23
chore: Support opening *.camel-k.yaml and *.pipe.yaml

igarashitm:
Fixes: #455


apupier (changes requested):
- it is already possible to open all *.yaml via right-click. In theory contextual menu should be provided even without activation
- *.camel-k.yaml and *.pipe.yaml are not standard naming from Camel community project. I do not think that we should list them directly.

PR #471, opened 2024-01-30
Fix Kaoto editor in vscode.dev web environment

djelinek:
Screenshot 2024-01-30 at 11 21 31

### DIFF
--- a/src/extension/extensionWeb.ts
+++ b/src/extension/extensionWeb.ts
@@ -20,6 +20,7 @@ import { EditorEnvelopeLocator, EnvelopeContentType, EnvelopeMapping } from "@ki
 import { I18n } from "@kie-tools-core/i18n/dist/core";
 import * as KogitoVsCode from "@kie-tools-core/vscode-extension/dist";
 import * as vscode from "vscode";
+import { KAOTO_FILE_PATH_GLOB } from "./helpers";
 
 let backendProxy: VsCodeBackendProxy;
 
@@ -36,7 +37,7 @@ export async function activate(context: vscode.ExtensionContext) {
     editorEnvelopeLocator: new EditorEnvelopeLocator("vscode", [
       new EnvelopeMapping({
         type: "kaoto",
-        filePathGlob: "**/*.kaoto.+(yml|yaml)",
+        filePathGlob: KAOTO_FILE_PATH_GLOB,
         resourcesPathPrefix: "dist/webview/editors/kaoto",
         envelopeContent: {
           type: EnvelopeContentType.PATH,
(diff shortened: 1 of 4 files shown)


src/extension/extensionWeb.ts
```diff
@@ -36,7 +36,7 @@ export async function activate(context: vscode.ExtensionContext) {
     editorEnvelopeLocator: new EditorEnvelopeLocator("vscode", [
       new EnvelopeMapping({
         type: "kaoto",
-        filePathGlob: "**/*.kaoto.+(yml|yaml)",
+        filePathGlob: "**/**(.+(kaoto|camel|kamelet)).+(yml|yaml)",
```
apupier, line 39: might be a good idea to factorize with a constant with the valeu used in extensions.ts

apupier (approved):

apupier (changes requested):
I think we should add a note in the changelog

src/extension/extensionWeb.ts
lordrip: If possible, can we make a super simple test that shows what strings are matched with this regexp?

apupier: this would depend on the regexp implementation used by the multiplying architecture which is not part of this codebase.

It is indirectly tested via UI tests that are checking that the files are opened with kaoto editor or not

apupier (approved):

apupier (changes requested):
oups compilation is no more working.

I think it is due to using a shared onstant and it is introducing more dependencies.
So I guess we need to put the constant in a specific file.

If nto working it would require change of webpack config, I think that it is not worthy to investigate on that side for vs code web dev, in this case I woudl suggest to continue to duplicate the regexp

src/extension/extensionWeb.ts
djelinek: > might be a good idea to factorize with a constant with the valeu used in extensions.ts

cannot be as part of extension.ts, so I have provided in separate file

djelinek:
> oups compilation is no more working.
> 
> I think it is due to using a shared onstant and it is introducing more dependencies. So I guess we need to put the constant in a specific file.
> 
> If nto working it would require change of webpack config, I think that it is not worthy to investigate on that side for vs code web dev, in this case I woudl suggest to continue to duplicate the regexp

exactly

src/extension/helpers.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
```
apupier, line 2: when creating a new file we shoudl use the curren tyear

djelinek: lol.. catched by copy/paste monster 🙁 😃

djelinek: fixed

apupier (approved):

PR #489, opened 2024-02-14
chore: adapt UI tests to main branch having route container

apupier:
- `pf-topology__node__action-icon` is used for the route container actions too, not found a cleaner way for now to pick the second entry in the list
- use replace instead of insert so that the test can work with both released and main branch of kaoto. We do not care in this test where the node is added
- the side panel needs to be awaited to be displayed on all OS now (which seems more coherent)
- added an ugly sleep, not found another way to wait for panel to be closed correctly. It seems like the close button is not ready immediately even if displayed, unless it is it due to animation and the click is sent to the wrong place?
- ignore extension recommendation: it avoids having a notification popup on Windows for Windows Subsystem Linux extension to interfere with the UI tests

### DIFF
--- a/it-tests/PropertyPanelLoading.test.ts
+++ b/it-tests/PropertyPanelLoading.test.ts
@@ -37,13 +37,16 @@ describe('Property panel loading test', function () {
     ), 5_000);
 
     const closeBtn = await driver.findElement(By.xpath("//button[@data-testid='close-side-bar']"));
-    if (process.platform === 'darwin') {
-      // from some reason this extra dynamic wait is needed on macOS
-      await driver.wait(async () => {
-        return await closeBtn.isDisplayed();
-      }, 5_000, 'Close button is not displayed!');
-    }
+    await driver.wait(async () => {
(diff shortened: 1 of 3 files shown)


it-tests/PropertyPanelLoading.test.ts
```diff
@@ -22,7 +22,7 @@ describe('Property panel loading test', function () {
     await editorView.closeAllEditors();
   });
 
-  it('Open "choice.camel.yaml" file and check property panel is loading and closing', async function () {
+  it.only('Open "choice.camel.yaml" file and check property panel is loading and closing', async function () {
```
djelinek, line 25: .only - needs to be removed before merge

djelinek (changes requested):

it-tests/PropertyPanelLoading.test.ts
apupier: :face_with_spiral_eyes:

djelinek (approved):

PR #518, opened 2024-03-06
feat: open pipe files by default

hdamarcelo:


apupier (changes requested):
* the readme needs to be updated: https://github.com/KaotoIO/vscode-kaoto/blob/33071bcdbc2da5ed136d4829bfd4d2416fbc221d/README.md?plain=1#L34-L35
* Please add a note in the changelog

djelinek (approved):
+1 after changelog is also updated

apupier (approved):

PR #583, opened 2024-05-27
changelog: provide information for telemetry updates which triggered a

apupier:
new release


apupier:
failure with main branch of kaoto is worked on in a separate issue

PR #596, opened 2024-06-25
Provide setting to specify an URL for a Catalog set

apupier:
part of #1204


lordrip:
I took these commits and added them here: https://github.com/KaotoIO/vscode-kaoto/pull/605

PR #605, opened 2024-07-08
feat(settings): Provide VSCode setting for the `catalogUrl`

lordrip:
### Context
Add a ´catalogUrl´ setting to point to a specific kaoto catalog version

### Note
In this PR we're updating the ´@kaoto/kaoto´ version to ´2.1.0-RC3´ so we can handle the case of an empty `catalogUrl`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.2.0
 
 - Update project homepage in the marketplace to [kaoto.io](https://kaoto.io)
+- Setting to provide custom set of Catalog used by the Kaoto editor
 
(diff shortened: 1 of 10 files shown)


djelinek:
Currently there is just kick-off test which set the custom catalog url using settings and then check the number of items is higher then 1 (because 1 item per each runtime should be the default value when no catalog is used)

**TODO:** in next iteration it would be nice to add test which will 
- set the catalog using  settings
- pick some different version of catalog in katoto editor
- and check that eg some components are now available

djelinek (approved):

PR #611, opened 2024-07-11
docs: change kaoto-next mentions to only kaoto

djelinek:


CONTRIBUTING.md
```diff
@@ -133,15 +133,15 @@ It is launching UI tests. Beware that it can take several minutes to start. Stay
...
-  * :exclamation: update `dependencies` section, update`"@kaoto-next/ui": "<version>"` to next version of publish Kaoto UI package
+  * :exclamation: update `dependencies` section, update`"@kaoto/kaoto": "<version>"` to next version of publish Kaoto UI package
   * :warning: update `resolutions` and `dependencies` section
 
   ```json
     // dependencies
-    // eg. for "@kaoto-next/ui": "2.0.0-TP2"
-    "@kaoto-next/ui": "2.0.0-TP2",
+    // eg. for "@kaoto/kaoto": "2.0.0-TP2"
+    "@kaoto/kaoto": "2.0.0-TP2",
```
lhein, line 144: should we still point to a TP version?

djelinek: in this place and the meaning of exact version is not crucial but I agree it will look better and @kaoto/kaoto was release as 2.0.0 so I will also update the version 🙂 

Thank you for spotting that Lars!

lhein (approved):

PR #622, opened 2024-07-24
feat(vscode): Provide SettingsAdapter for VSCode

lordrip:
### Context
This commit adds a new `getVSCodeKaotoSettings` method to the `VSCodeKaotoEditorChannelApi` class to build the Settings model.

It also deprecates the existing `getCatalogURL` method as the new way to go is using the settings adapter from now on.

### Note
This PR needs to be merged after: https://github.com/KaotoIO/kaoto/pull/1264

relates: https://issues.redhat.com/browse/KTO-441
relates: https://github.com/KaotoIO/kaoto/pull/1221
relates: https://github.com/KaotoIO/kaoto/pull/1264

### DIFF
--- a/src/webview/VSCodeKaotoEditorChannelApi.ts
+++ b/src/webview/VSCodeKaotoEditorChannelApi.ts
@@ -1,8 +1,19 @@
+import { KaotoEditorChannelApi } from '@kaoto/kaoto';
+import { ISettingsModel, NodeLabelType, SettingsModel } from '@kaoto/kaoto/models';
 import { DefaultVsCodeKieEditorChannelApiImpl } from '@kie-tools-core/vscode-extension/dist/DefaultVsCodeKieEditorChannelApiImpl';
 import * as vscode from 'vscode';
 
-export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelApiImpl {
-    async getCatalogURL(): Promise<string | undefined> {
-        return await vscode.workspace.getConfiguration('kaoto').get('catalog.url');
-    }
+export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelApiImpl implements KaotoEditorChannelApi {
+  async getCatalogURL(): Promise<string | undefined> {
+    return await vscode.workspace.getConfiguration('kaoto').get('catalog.url');
+  }
+
+  async getVSCodeKaotoSettings(): Promise<ISettingsModel> {
+    const settingsModel: ISettingsModel = {
+      catalogUrl: vscode.workspace.getConfiguration('kaoto').get<string | null>('catalog.url') ?? '',
+      nodeLabel: NodeLabelType.Description,
+    };
+
+    return new SettingsModel(settingsModel);
+  }
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -20,7 +20,7 @@
 	  "downlevelIteration": true,
 	  "skipLibCheck": true,
 	  "jsx": "react-jsx",
-	  "moduleResolution":"Node",
+	  "moduleResolution":"Bundler",
 	  "esModuleInterop": true
 	}
 }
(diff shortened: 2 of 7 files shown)


src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -1,8 +1,20 @@
...
+  extends DefaultVsCodeKieEditorChannelApiImpl
+  implements KaotoEditorChannelApi
+{
+  async getCatalogURL(): Promise<string | undefined> {
+    return await vscode.workspace.getConfiguration('kaoto').get('catalog.url');
+  }
+
+  async getVSCodeKaotoSettings(): Promise<ISettingsModel> {
+    const catalogUrl =
+      vscode.workspace.getConfiguration('kaoto').get<string | null>('catalog.url') ?? '';
```
apupier, line 16: if I understand well, we now return an empty string in case the setting is not set (so user let to its default)?
Is it fine to no more distinguish between the 2 states? (I guess that in the case of catalogUrl it is fine)

src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -1,8 +1,20 @@
...
+
+    return new SettingsModel({ catalogUrl });
```
apupier, line 18: Should it be needed to specify that the value is assigned to the catalogUrl is we do not want to break the compilation when another value will be added?

(it will be more explicit too)

```
return new SettingsModel({ catalogUrl: catalogUrl });
```

src/webview/VSCodeKaotoEditorChannelApi.ts
apupier, line 16: from a style point of view, you can really write on the same line the1 07 characters. I think almost everybody have monitor  which can handle a lot more than 80 characters nowadays

lordrip: Yes, so far, the `catalogUrl` from VSCode can be `null` if the user hasn't set it or a `string` that could be empty or not, but on the Kaoto side, `catalogUrl` can only be a string (no null) hence we return a string here.

lordrip: Sure, it probably picked the settings from Kaoto workspace, I'll make it one line

src/webview/VSCodeKaotoEditorChannelApi.ts
lordrip: This `{ catalogUrl: catalogUrl }` is the same as `{ catalogUrl  }`, also the latter is a well-known JS pattern, would you be open to initialize the settings object and then provide it to the `SettingsModel` constructor?

```javascript
const settingsModel: ISettingsModel = {
  catalogUrl: vscode.workspace.getConfiguration('kaoto').get<string | null>('catalog.url') ?? '',
};

```

tsconfig.json
```diff
@@ -20,7 +20,7 @@
 	  "downlevelIteration": true,
 	  "skipLibCheck": true,
 	  "jsx": "react-jsx",
-	  "moduleResolution":"Node",
+	  "moduleResolution":"Bundler",
```
lordrip, line 23: This is needed to be able to use the new export from kaoto: `@kaoto/kaoto/models`

djelinek (approved):

djelinek (approved):
I am wondering should we mention the bump of @kaoto/kaoto to 2.2.0-RC3 in changelog and readme to have main branch up to date?

lordrip:
> I am wondering should we mention the bump of @kaoto/kaoto to 2.2.0-RC3 in changelog and readme to have main branch up to date?

I'll take care :+1:

lordrip:
> > I am wondering should we mention the bump of @kaoto/kaoto to 2.2.0-RC3 in changelog and readme to have main branch up to date?
> 
> I'll take care 👍

Updated the CHANGELOG.md and the READMEmd

PR #633, opened 2024-08-09
feat(build): Update Kaoto to 2.2.0-RC4

lordrip:
### Context
Since other consumers need to use `@kaoto/kaoto`, we need to remove as many dependencies as possible from Kaoto itself and provide them as `peerDependencies`.

This PR updates `@kaoto/kaoto` to `2.2.0-RC4` and provides the missing dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,14 +30,23 @@
     "test:it:clean": "rimraf ./test-resources && rimraf ./out && rimraf *.vsix"
   },
   "dependencies": {
-    "@kaoto/kaoto": "2.2.0-RC3",
+    "@kaoto/kaoto": "2.2.0-RC4",
     "@kie-tools-core/backend": "0.32.0",
     "@kie-tools-core/editor": "0.32.0",
     "@kie-tools-core/i18n": "0.32.0",
     "@kie-tools-core/vscode-extension": "0.32.0",
+    "@patternfly/patternfly": "5.3.1",
+    "@patternfly/react-code-editor": "5.3.4",
+    "@patternfly/react-core": "5.3.4",
+    "@patternfly/react-icons": "5.3.2",
+    "@patternfly/react-table": "5.3.4",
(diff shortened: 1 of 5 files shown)


apupier (changes requested):
- the versions mentioned in readme must be updated too https://github.com/KaotoIO/vscode-kaoto/blob/a8e655f3c4a68ef87d2a43e3acc5e97cac74b84e/README.md?plain=1#L7 https://github.com/KaotoIO/vscode-kaoto/blob/a8e655f3c4a68ef87d2a43e3acc5e97cac74b84e/README.md?plain=1#L46
- What an awful pattern having to define extra-dependencies to make one of them working. it will add a bunch of maintenance burden to ensure the synchronization of versions

package.json
```diff
@@ -30,14 +30,23 @@
...
+    "@patternfly/react-icons": "5.3.2",
+    "@patternfly/react-table": "5.3.4",
+    "@patternfly/react-topology": "5.4.0-prerelease.13",
     "@redhat-developer/vscode-redhat-telemetry": "^0.8.0",
+    "monaco-editor": "^0.50.0",
+    "monaco-yaml": "^5.1.1",
     "react": "18.2.0",
-    "react-dom": "18.2.0"
+    "react-dom": "18.2.0",
+    "react-monaco-editor": "^0.56.0"
```
apupier, line 49: why these dependencies are added here and not in the resolutions sections? What's the difference with the patternfly dependencies?

lordrip: there are two consumers of patternfly dependencies:
1. Kaoto
2. KIE tools

From vscode-kaoto side, we need to provide those dependencies, hence we add them in the `dependencies` section. Unfortunately, KIE tools require a different version of the dependencies, so in this case, we also add the patternfly dependencies to the resolution field, so all packages inside vscode-kaoto can use the same exact patternfly version and avoid the duplicity.

apupier (approved):

PR #640, opened 2024-08-20
Fix getCatalogUrl missing API in web version

apupier:
fixes #639

PR #641, opened 2024-08-20
482 fix open with kaoto editor in web version

apupier:
based on https://github.com/KaotoIO/vscode-kaoto/pull/640

### DIFF
--- a/src/extension/extensionWeb.ts
+++ b/src/extension/extensionWeb.ts
@@ -31,7 +31,7 @@ export async function activate(context: vscode.ExtensionContext) {
   const backendI18n = new I18n(backendI18nDefaults, backendI18nDictionaries, vscode.env.language);
   backendProxy = new VsCodeBackendProxy(context, backendI18n);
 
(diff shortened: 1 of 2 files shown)


src/extension/extensionWeb.ts
```diff
@@ -50,6 +50,17 @@ export async function activate(context: vscode.ExtensionContext) {
     backendProxy: backendProxy,
   });
 
+  vscode.commands.registerCommand('kaoto.open.textualeditor', async() => {
+    if (kieEditorStore.activeEditor !== undefined) {
+      const doc = await vscode.workspace.openTextDocument(kieEditorStore.activeEditor?.document.document.uri);
```
djelinek, line 55: it is really ".document.document" syntax? it looks a bit strange, so I am just asking..

djelinek (approved):

src/extension/extensionWeb.ts
apupier: yes, I confirm it is really this API

PR #653, opened 2024-09-12
chore(API): Implement FilePreferences methods

lordrip:
This is the dummy implementation for the upcoming API required by Kaoto

### DIFF
--- a/src/webview/VSCodeKaotoEditorChannelApi.ts
+++ b/src/webview/VSCodeKaotoEditorChannelApi.ts
@@ -19,4 +19,14 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
 
     return new SettingsModel(settingsModel);
   }
+
+  async getFilePreferences<T>(key: string): Promise<T> {
+    vscode.window.showInformationMessage('Method not implemented.');
+    return {} as T;
+  }
+
+  async setFilePreferences(key: string): Promise<void> {
+    vscode.window.showInformationMessage('Method not implemented.');
+  }
+
 }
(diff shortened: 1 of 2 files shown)


lhein (approved):

src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -19,4 +19,14 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
 
     return new SettingsModel(settingsModel);
   }
+
+  async getFilePreferences<T>(key: string): Promise<T> {
```
apupier, line 23: I'm wondering if we do not need the corresponding route filepath as parameter to be able to find the correct .kaoto file?
Or we have access to it because there is one channel API per editor instance and can access it?

lordrip: I guess since the `VSCodeKaotoEditorChannelApi` class extends `DefaultVsCodeKieEditorChannelApiImpl`, we could this information and more in one of the attached services.

Here's the `DefaultVsCodeKieEditorChannelApiImpl` class constructor

```ts
export class DefaultVsCodeKieEditorChannelApiImpl implements KogitoEditorChannelApi, JavaCodeCompletionChannelApi {
  constructor(
    private readonly editor: VsCodeKieEditorController,
    private readonly resourceContentService: ResourceContentService,
    private readonly workspaceApi: WorkspaceChannelApi,
    private readonly backendProxy: BackendProxy,
    private readonly notificationsApi: NotificationsChannelApi,
    private readonly javaCodeCompletionApi: JavaCodeCompletionApi,
    private readonly viewType: string,
    private readonly i18n: I18n<VsCodeI18n>
  ) {}
```

lordrip: Would you prefer to have a `path` parameter here?

lordrip: Oh, that's not good, all members are `private` :facepalm:

lordrip:
Closing this PR for now to repurpose the API interface first

PR #656, opened 2024-09-13
New Channel API to manipulate metadata and content of other files

apupier:


src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -19,4 +45,27 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
...
+
+    /**
+   * Store metadata in the nearest .kaoto file found. If none found, create the file at workspace root.
+   */
+    async setMetadata<T>(key: string, value: T | undefined): Promise<void> {
+
+    }
+
+    private async findKaotoMetadataFile(camelRouteFileUri: vscode.Uri): Promise<vscode.Uri | undefined> {
+      // TODO really rtrieve the file by looking up the path
```
lordrip, line 69:
```suggestion
      // TODO really retrieve the file by looking up the path
```

src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -19,4 +45,27 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
...
   }
+
+  /**
+   * Provide metadata stored in the nearest .kaoto file found
+   */
+  async getMetadata<T>(key: string): Promise<T | undefined> {
+    const kaotoMetadatafile: vscode.Uri | undefined = await this.findKaotoMetadataFile(this.currentEditedDocument.uri);
+    if (kaotoMetadatafile != undefined) {
+      const kaotoMetadataFileContent = new TextDecoder().decode(await vscode.workspace.fs.readFile(kaotoMetadatafile));
+      return JSON.parse(kaotoMetadataFileContent).get(key);
```
lordrip, line 56: I think this will :boom:, as `JSON.parse` could return an object, a number, a string or throw an exception.

My suggestion would be to create a simple class that opens the given path and ensure the content is an object and not anything else.

This class will also implement getters and setters for interacting with it.

```ts
class MetadataFileHandler {
  protected metadata: Record<string, unknown> = {};

  open(path: string) {
    try {
      const result = JSON.parse(kaotoMetadataFileContent);
      if (typeof localMetadata === 'object' && !Array.isArray(localMetadata)) {
        this.metadata = result;
      }

      throw new Error('Invalid metadata file');
    } catch {
      // Handle parsing errors here
      this.metadata = {};
    }
  }

  get<T>(key: string): T {
    return this.metadata[key] as T;
  }

  set<T>(key: string, value: T) {
    this.metadata[key] = value;
  }
}
```

src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -19,4 +45,27 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
 
     return new SettingsModel(settingsModel);
   }
+
+  /**
+   * Provide metadata stored in the nearest .kaoto file found
+   */
+  async getMetadata<T>(key: string): Promise<T | undefined> {
+    const kaotoMetadatafile: vscode.Uri | undefined = await this.findKaotoMetadataFile(this.currentEditedDocument.uri);
+    if (kaotoMetadatafile != undefined) {
```
lordrip, line 54:
```suggestion
    if (kaotoMetadatafile !== undefined) {
```

src/webview/VSCodeKaotoEditorChannelApi.ts
apupier: the content of the .kaoto file is what we are creating, so if it explodes it means that it is not our file

lordrip: ...or it is but is corrupted, the problematic part is not just the :boom:, but rather what's gonna happen afterward, are we gonna end up in an unstable situation that will require the editor to be reopened?

As you said at the beginning, users might choose not commiting the metadata file, in that case, I don't think we should leave the editor in an unusable state, but rather accommodate for the possibility it could happen.

apupier: this is up to our choice of implementation. Are we expeting to bubblingup exception so tha they can be handled property at the editor level? or we return undefined when we cannot find the metadata for whatever reasons. And reporting to users wil lthen be specific to client and not handled centrally at kaoto editor level

lordrip: > or we return undefined when we cannot find the metadata for whatever reasons.

Probably this one, the point that I tried to make was more toward avoiding uncontrolled exceptions happening, so we don't end up with an inconsistent state, this could happen in Kaoto (we're gonna handle it at the Kaoto side) or it could be at the MA side, that's why I thought of mentioning that we better avoid throwing uncontrolled exceptions.

igarashitm (approved):
:+1:

PR #659, opened 2024-09-16
Introduce unit/integration tests

### LINKED ISSUE #658: Introduce unit/integration tests
currently there are only end to end UI tests. it was enough so far as only integration points with Kaoto and few UI endpoint contribution were made. With more code inserted in VS Code to implement the Channel APIs for metadata and content of file access from kaoto through the Channel API, being able to test at unit level is a must have.

technical notes:
* Our other VS Code extensions and official examples are using Mocha (I think that Kaoto is using Jest)
* Surely some additional configurations to provide compared to our other extensions as it is the only one which is webpacked and using yarn.

apupier:
base configuration coming from:
https://github.com/microsoft/vscode-extension-samples/tree/main/helloworld-test-sample

fixes #658


.vscode-test.mjs
```diff
@@ -0,0 +1,12 @@
...
+export default defineConfig({
+	files: 'out/test/**/*.test.js',
+	workspaceFolder: './test Fixture with speci@l chars',
+	mocha: {
+		ui: 'tdd',
+		color: true,
+		timeout: 100000,
+		reporter: 'mocha-jenkins-reporter'
+	}
+});
```
apupier, line 12: Interesting point: this could replace several of our runTests class in our other VS Code extensions.

.gitignore
```diff
@@ -15,6 +15,7 @@ dist
 # testing
 /coverage
 .vscode-test-web
+.vscode-test
```
djelinek, line 18: I would siggest to add .vscode-test into .vscodeignore to not bundle into .vsix package

apupier: I was lucky, it is already set in .vscodeignore https://github.com/KaotoIO/vscode-kaoto/blob/1393a8a8d3351ee58e920dcabebcb9eb5451260f/.vscodeignore#L2

djelinek (approved):

PR #663, opened 2024-09-19
Ensure dset 3.1.4+ is used

apupier:
Previous version comes through transitive dependencies with a version containing CVE


apupier:
required to specify the resolution until vscode-redhat-telemetry has not upgraded their depdency and doa new release https://github.com/redhat-developer/vscode-redhat-telemetry/pull/55

PR #664, opened 2024-09-19
Upgrade embedded Kaoto to 2.2.0-RC6

apupier:


apupier:
@lordrip Do you see a dependency version for which I could have missed the alignment?

lordrip:
> @lordrip Do you see a dependency version for which I could have missed the alignment?

No, so far everything looks good at VSCode side @apupier. :+1:

PR #698, opened 2024-10-24
Change implementation of VS Code Channel editor API to have it

apupier:
compatible with Maven based projects

basically looking for classpath root src/main/resources if there is some for the xsl file. And using a relative path to the .kaoto file for the xsd file
demoWithMavenBased

### DIFF
--- a/src/webview/VSCodeKaotoEditorChannelApi.ts
+++ b/src/webview/VSCodeKaotoEditorChannelApi.ts
@@ -12,6 +12,7 @@ import * as vscode from 'vscode';
 import { VsCodeKieEditorCustomDocument } from '@kie-tools-core/vscode-extension/dist/VsCodeKieEditorCustomDocument';
 import * as path from 'path';
 import { logInKaotoOutputChannel } from './../KaotoOutputChannelManager';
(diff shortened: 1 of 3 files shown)


src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -94,51 +88,54 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
...
 
+  private findClasspathRoot() {
+    const camelRoutePath = this.currentEditedDocument.uri.fsPath;
+    const folderClasspathPattern = `src${path.sep}main${path.sep}resources`;
+    const indexOfClasspath = camelRoutePath.lastIndexOf(folderClasspathPattern);
+    let classpathRoot: string;
+    if (indexOfClasspath !== -1) {
+      classpathRoot = camelRoutePath.substring(0, indexOfClasspath + folderClasspathPattern.length);
+    } else {
+      // In non Maven based project, we consider the folder at same level than Camel route to bethe classpath route
```
lordrip, line 112:
```suggestion
      // In non Maven based project, we consider the folder at same level than Camel route to be the classpath route
```

djelinek (approved):

PR #705, opened 2024-10-31
Temporary workaround to get ResourceContent

apupier:
This is an ugly workaround that could lead to path clashes. 2 different APIs must be provided to retrieve content, one relative to classpath root and the other relative to .kaoto metadata file, reported https://issues.redhat.com/browse/FUSETOOLS2-2505

### DIFF
--- a/src/webview/VSCodeKaotoEditorChannelApi.ts
+++ b/src/webview/VSCodeKaotoEditorChannelApi.ts
@@ -95,10 +95,21 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
       const targetFile = path.resolve(classpathRoot, relativePath);
       return new TextDecoder().decode(await vscode.workspace.fs.readFile(vscode.Uri.file(targetFile)));
     } catch (ex) {
-      const errorMessage= `Cannot retrieve content of ${relativePath} relatively to ${classpathRoot}`;
-      vscode.window.showErrorMessage(errorMessage);
+      // TODO: this is an ugly workaround that could lead to path clashes. 2 different APIs must be provided to retrieve content, one relative to classpath and the other relative to .kaoto metadata file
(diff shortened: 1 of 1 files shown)


src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -95,10 +95,19 @@ export class VSCodeKaotoEditorChannelApi extends DefaultVsCodeKieEditorChannelAp
...
     } catch (ex) {
-      const errorMessage= `Cannot retrieve content of ${relativePath} relatively to ${classpathRoot}`;
-      vscode.window.showErrorMessage(errorMessage);
-      logInKaotoOutputChannel(errorMessage, ex);
-      return undefined;
+      // TODO: this is an ugly workaround that could lead to path clashes. 2 different APis must be provided to retrieve content, one relative to classpath and the other relative to .kaoto metadata file
+      let kaotoMetadataFile = await this.findExistingKaotoMetadataFile(this.currentEditedDocument.uri);
+      try {
+        if (kaotoMetadataFile !== undefined) {
+          const targetFile = path.resolve(path.dirname(kaotoMetadataFile?.fsPath), relativePath);
```
lordrip, line 102: Since the `kaotoMetadataFile` is validated in the `if`statement, I think we can spare the optional chaining.
```suggestion
          const targetFile = path.resolve(path.dirname(kaotoMetadataFile.fsPath), relativePath);
```

Other than that, wouldn't be interesting to also log this exception? Like an `info` log of some sort explaining that we tried to find the file relative to something and it wasn't found, so now we're falling back to relative to the `.kaoto` file.

apupier:
not enough it creates another duplicated xsd file in a subfolder:
image

apupier:
> not enough it creates another duplicated xsd file in a subfolder:
> image

This specific issue is due to an issue in datamapper itself, see https://github.com/KaotoIO/kaoto-datamapper-integration/issues/194

igarashitm (approved):

PR #709, opened 2024-11-15
fix(Settings): Use default Settings from Kaoto

lordrip:
### Context
From the Kaoto side, we have default settings in the
[settings.model.ts](https://github.com/KaotoIO/kaoto/blob/9ee3368e0e488853f61a84fa4e16bedadc801fda/packages/ui/src/models/settings/settings.model.ts#L23-L25) file.

### Changes
This commit uses that and only overrides what needs to be changed in terms of the `catalogUrl` and the `nodeLabel`.

### Notes
Once the VS Code setting for the new property `nodeToolbarTrigger` is merged, we can add it also to the VS Code settings object.

### DIFF
--- a/it-tests/BasicFlow.test.ts
+++ b/it-tests/BasicFlow.test.ts
@@ -151,10 +151,11 @@ async function createNewRoute(driver: WebDriver) {
 
 async function addActiveMQStep(driver: WebDriver) {
   await driver.wait(
-    until.elementLocated(By.xpath('//*[@data-type="node"]//*[@class="pf-topology__node__action-icon"]'))
+    until.elementLocated(By.css('g[data-testid^="custom-node__timer"]'))
   );
-  const threeDotsIconOfOneOfTheSteps = (await driver.findElements(By.xpath('//*[@data-type="node"]//*[@class="pf-topology__node__action-icon"]')))[1];
-  await threeDotsIconOfOneOfTheSteps.click();
+
+  const canvasNode = await driver.findElement(By.css('g[data-testid^="custom-node__timer"]'));
+  await driver.actions().contextClick(canvasNode).perform();
 
   await driver.wait(
     until.elementLocated(By.className('pf-v5-c-dropdown pf-m-expanded'))
--- a/it-tests/settings/NodeLabelSettings.test.ts
+++ b/it-tests/settings/NodeLabelSettings.test.ts
@@ -14,8 +14,8 @@ describe('User Settings', function () {
 
     const locators = {
         TimerComponent: {
-            timer: By.xpath(`//\*[name()='g' and starts-with(@data-id,'timer')]`),
-            label: By.xpath(`//\*[name()='g' and starts-with(@class,'pf-topology__node__label')]`)
+            timer: `g[data-id^='timer'][data-kind='node']`,
+            label: `.custom-node__label`,
         }
     }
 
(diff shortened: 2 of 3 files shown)


apupier (approved):
it is fixing the compilation with main branch of kaoto
Several tests are broken with main branch of Kaoto but it sounds unrelated and could be tackled in another PR

lordrip:
> it is fixing the compilation with main branch of kaoto Several tests are broken with main branch of Kaoto but it sounds unrelated and could be tackled in another PR

The failures are likely due to the changes in the canvas nodes, I'll check

it-tests/BasicFlow.test.ts
```diff
@@ -151,10 +151,11 @@ async function createNewRoute(driver: WebDriver) {
...
 async function addActiveMQStep(driver: WebDriver) {
   await driver.wait(
-    until.elementLocated(By.xpath('//*[@data-type="node"]//*[@class="pf-topology__node__action-icon"]'))
+    until.elementLocated(By.css('g[data-testid^="custom-node__timer"]'))
   );
-  const threeDotsIconOfOneOfTheSteps = (await driver.findElements(By.xpath('//*[@data-type="node"]//*[@class="pf-topology__node__action-icon"]')))[1];
-  await threeDotsIconOfOneOfTheSteps.click();
+
+  const canvasNode = await driver.findElement(By.css('g[data-testid^="custom-node__timer"]'));
+  driver.actions().contextClick(canvasNode).perform();
```
lordrip, line 158: Since the 3-dots button is no longer there, we're triggering a right-click event over the node to get the same menu.

it-tests/settings/NodeLabelSettings.test.ts
```diff
@@ -14,8 +14,8 @@ describe('User Settings', function () {
 
     const locators = {
         TimerComponent: {
-            timer: By.xpath(`//\*[name()='g' and starts-with(@data-id,'timer')]`),
-            label: By.xpath(`//\*[name()='g' and starts-with(@class,'pf-topology__node__label')]`)
+            timer: `g[data-id^='timer'][data-kind='node']`,
+            label: `.custom-node__label`,
```
lordrip, line 18: This selector changed because there's no more topology label but rather a standard text.

it-tests/settings/NodeLabelSettings.test.ts
```diff
@@ -64,7 +64,7 @@ describe('User Settings', function () {
     it(`Check 'id' Node Label is used instead of default 'description'`, async function () {
         this.timeout(60_000);
 
-        const timer = await driver.findElement(locators.TimerComponent.timer).findElement(locators.TimerComponent.label);
+        const timer = await driver.findElement(By.css(`${locators.TimerComponent.timer} ${locators.TimerComponent.label}`));
```
lordrip, line 67: Semantically, this is not exactly the same as before.

Before, we were locating a node and from that point, we searched for another one within. Now, we're targeting the label that is inside of the node.

The issue is that for some reason, topology now creates invisible nodes, potentially as references inside of the groups, so it means the test library finds 2 nodes, instead of one, and the first one is empty, therefore, the location fails.

it-tests/BasicFlow.test.ts
djelinek: just a small one for the stabillity.. is the "driver.actions...perform()" a floating Promise? I think we could missing 'await' here

lordrip: I don't know, I'll check

lordrip: Alrighty, I was missing the `await`, `perform` is certainly a `Promise`, thanks for the catch @djelinek

Issue #747, opened 2024-12-22
Datamapper unable to import XSD on Windows


chiragsanghavi:
* I have created a yaml route using kaoto
* Added kaoto-datamapper in the route
* copied xsd into my folder that has both .kaoto file and yaml file
* in the datamapper configuration - want to attach schema in source and target
* while trying to import schema - it says `Cannot retrieve content of ..\..\c:\myfolder\project\schema.xsd relatively to the classpath root c:\myfolder\project neither the .kaoto metadata file file:///c%3A/myfolder/project/.kaoto`

lhein:
Thanks for reporting the issue. We will check what's wrong.

lhein:
Thanks for reporting back! Glad that it works for you now.

PR #748, opened 2024-12-23
First iteration UI test adding a Kaoto datamapper step

apupier:
in this iteration only checking that it is available in the catalog and that the step is added to the canvas

preparatory work to provide non-regression test for #747


apupier:
let's review directly https://github.com/KaotoIO/vscode-kaoto/pull/749

PR #759, opened 2025-01-08
Precise that Kaoto editors must be closed and reopened on settings to be effective

apupier:


pospisilf:
@apupier I think using capital letter in Kaoto would look better, what do you think?

pospisilf (commented):
I think using capital letter in Kaoto would look better, what do you think?

pospisilf (approved):

PR #772, opened 2025-01-16
Feat/update kaoto 2.5.0-M2

apupier:


webpack.config.js
```diff
@@ -118,7 +118,7 @@ const commonConfig = (env) => {
       alias: {
         "react": path.resolve('./node_modules/react'),
         "react-dom": path.resolve('./node_modules/react-dom'),
-        "@patternfly/react-core": path.resolve('./node_modules/@patternfly/react-core')
+        // "@patternfly/react-core": path.resolve('./node_modules/@patternfly/react-core')
```
lordrip, line 121: I forgot to remove this line from the previous PR, when we have the chance, let's :boom: it

webpack.config.js
```diff
@@ -118,7 +118,6 @@ const commonConfig = (env) => {
       alias: {
         "react": path.resolve('./node_modules/react'),
         "react-dom": path.resolve('./node_modules/react-dom'),
-        "@patternfly/react-core": path.resolve('./node_modules/@patternfly/react-core')
```
apupier, line 121: @lordrip What is the impact of removing it? (apart from the good news that it allows compilation despite kie using patternfly v4)

lordrip (approved):

webpack.config.js
apupier: if I understand well, now if kie-tools-core is using some patternfly component they will use Patternfly v4 for it, right?
(I would need to search in which case they are used and if it can occur in our context to check that it is working)

lordrip: Yes, it will and it's happening already for the components used by the hotkeys panel (which we don't see) and the initial loading distractor

apupier:
on hold: discovered that it breaks the Drag'n'drop (only in VS Code...)

apupier:
> on hold: discovered that it breaks the Drag'n'drop (only in VS Code...)

will be fixed by https://github.com/KaotoIO/kaoto/pull/1922

apupier:
some config node cannot be loaded, here the datamapper:

image

PR #776, opened 2025-01-27
Add Camel extensions as dependencies

apupier:
It allows to have several general Camel features such as Run/debug/Create/Export for users which are installing VS Code Kaoto extension only.


apupier:
closing as it was decided to have VS Code Kaoto as fully standalone withotu external dependencies. SMeans that several features will be moved from the other extensions to this one

PR #793, opened 2025-02-17
feat: add logic to log into Kaoto Output Channel

djelinek:
a bit reworked logger into Kaoto Output Channel which will be then reused for sharing information what is happening when butons, views, etc are invoking some actions

Screenshot 2025-02-17 at 14 56 13


src/extension/KaotoOutputChannel.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright 2024 Red Hat, Inc. and/or its affiliates.
```
apupier, line 2: maybe time to update to 2025

src/extension/KaotoOutputChannel.ts
```diff
@@ -0,0 +1,54 @@
...
+        const timestamp = new Date().toUTCString();
+        this.getInstance().appendLine(`[${timestamp}] [${level}] ${message}`);
+    }
+
+    // Dispose the Output Channel (mainly when the extension is deactivated)
+    public static dispose(): void {
+        if (KaotoOutputChannel.instance) {
+            KaotoOutputChannel.instance.dispose();
+        }
+    }
```
apupier, line 53: I think we need to also set back `KaotoOutputChannel.instance` to `undefined`. Otherwise when extensions are restarted, it will fail next time it tries to log something

djelinek: I was not able to simulate using `Reload Window` nor `Extensions: Refresh` nor disable/enable Kaoto extension

apupier: I needed to build the vsix and install it to go through activate/deactivate of the extension. (in dev mode the disable/enable is using the installed extension, not the one under development)
But just tried and it is working anyway. i do not understand why the `KaotoOutputChannel.instance` when disposed is not throwing an error the next time it is called

apupier (approved):

PR #795, opened 2025-02-19
feat: sort out user settings into sub categories

djelinek:
BEFORE:
Screenshot 2025-02-19 at 17 03 17

AFTER:
Screenshot 2025-02-19 at 16 55 17

### DIFF
--- a/package.json
+++ b/package.json
@@ -164,52 +164,65 @@
         }
       ]
     },
-    "configuration": {
-      "properties": {
-        "redhat.telemetry.enabled": {
-          "type": "boolean",
-          "default": null,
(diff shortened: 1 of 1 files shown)


apupier (approved):

lordrip:
@djelinek, do this affects how we read the configuration from Kaoto?

djelinek:
> @djelinek, does this affect how we read the configuration from Kaoto?

No, it doesn't. Settings ID's were not changed so that is why `VSCodeKaotoEditorChannelApi` and even UI tests do not need to be changed

apupier:
> does this affect how we read the configuration from Kaoto?

No, it doesn't. The property id, which is passed to Kaoto, remains the same.

lordrip (approved):

PR #798, opened 2025-02-24
feat: add New Camel File... button into Kaoto: Integrations view

djelinek:
Supports

- Routes, Pipes and Kamelets YAML DSL
- single workspace folder
- multi workspace folders

image

### DIFF
--- a/src/commands/AbstractNewCamelRouteCommand.ts
+++ b/src/commands/AbstractNewCamelRouteCommand.ts
@@ -1,0 +1,91 @@
+/**
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { commands, Uri, window, workspace, WorkspaceFolder } from "vscode";
+import { AbstractCamelCommand } from "./AbstractCamelCommand";
+import { KaotoOutputChannel } from "../extension/KaotoOutputChannel";
+
+export abstract class AbstractNewCamelRouteCommand extends AbstractCamelCommand {
+
+	protected fileNameInputPrompt = 'Please provide a name for the new file (without extension).';
+
+	protected async showInputBoxForFileName(targetFolder?: string): Promise<string> {
+		const input = await window.showInputBox({
+			prompt: this.fileNameInputPrompt,
+			placeHolder: this.camelDSL?.placeHolder ?? '',
+			validateInput: (fileName) => {
+				return this.validateCamelFileName(fileName ?? '', targetFolder);
+			},
+		});
+		return input ?? '';
+	}
+
+	protected async showDialogToPickFolder(defUri?: Uri): Promise<Uri | undefined> {
+		const selectedFolders = await window.showOpenDialog(
+			{
+				canSelectMany: false,
+				canSelectFolders: true,
+				canSelectFiles: false,
+				openLabel: 'Select',
+				title: 'Select a folder to create the file in. ESC to cancel a file creation.',
+				defaultUri: defUri || this.singleWorkspaceFolder?.uri
+			});
+		return selectedFolders ? selectedFolders[0] : undefined;
+	}
+
+	protected async showWorkspaceFolderPick(): Promise<WorkspaceFolder | undefined> {
+		if (workspace.workspaceFolders && workspace.workspaceFolders.length > 1) {
+			return await window.showWorkspaceFolderPick();
+		}
+		return undefined;
+	}
+
+	protected async showNoWorkspaceNotification(): Promise<void> {
+		const warningMessage = 'Missing workspace context! This action requires an open workspace. Please add a folder or reopen a project.';
+		KaotoOutputChannel.logWarning(warningMessage);
+		const selection = await window.showWarningMessage(warningMessage, 'Open Folder');
+		if (selection !== undefined) {
+			await commands.executeCommand('vscode.openFolder');
+		}
+	}
+
+	/**
+	 * Waits until a file for a given vscode.Uri exists.
+	 *
+	 * @param fileUri the file URI to check.
+	 * @param maxWaitMs maximum wait time in milliseconds (default: 5 seconds).
+	 * @param delayMs initial delay time in milliseconds (default: 100ms).
+	 * @returns a promise that resolves when the file exists or rejects when the timeout is reached.
+	 */
+	protected async waitForFileExists(fileUri: Uri, maxWaitMs: number = 5_000, delayMs: number = 100): Promise<void> {
+		return new Promise((resolve, reject) => {
+			const startTime = Date.now();
+			const checkFile = async () => {
+				try {
+					await workspace.fs.stat(fileUri);
+					resolve();
+				} catch {
+					if (Date.now() - startTime >= maxWaitMs) {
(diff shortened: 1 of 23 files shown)


djelinek:
It is marked as a `Draft` because UI tests are still in progress.. but feature is ready to test. 

May I ask you @apupier @lhein @lordrip to give a try locally when you have a couple of minutes please? (if it helps, here is a [link to VSIX](https://drive.google.com/file/d/1T1d1XuirfDFrpudzPNveYw3k6cNeaWJ-/view?usp=sharing))

djelinek:
FYI: 😱 I have spotted some issues in webExtension entrypoint.. but the regular one should be working as expected

src/commands/NewCamelKameletCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export class NewCamelKameletCommand extends AbstractNewCamelRouteCommand {
+
+	public static readonly ID_COMMAND_CAMEL_KAMELET_YAML = 'kaoto.camel.jbang.init.kamelet.yaml';
+	protected static readonly PROGRESS_NOTIFICATION_MESSAGE = 'Creating a new Kamelet file...';
+
+	private kameletType: string = '';
+
+	public async create(): Promise<void> {
+		const wsFolder = await this.showWorkspaceFolderPick();
```
apupier, line 30: When escaping the workspace folder selection, instead of aborting the command, it goes to next step and ask for a folder to choos

apupier (commented):
- given that we cannot create afile when no folde ris opened, we should either hide the "new file" action from the view or provide a message mentioning it. Currently ther eis avaliadation message on the second step about a "workspace folder undefined"

djelinek:
> FYI: 😱 I have spotted some issues in webExtension entrypoint.. but the regular one should be working as expected

it seems Web Extensions are not supporting Shell Execution for vscode.Tasks 🙁

 `VS Code can run ShellExecution tasks, such as when VS Code is run as a desktop application or when using one of the remote extensions, such as Dev Containers.`

src/commands/AbstractNewCamelRouteCommand.ts
```diff
@@ -0,0 +1,81 @@
...
+
+	/**
+	 * Waits until a file for a given vscode.Uri exists.
+	 *
+	 * @param fileUri the file URI to check.
+	 * @param maxWaitMs maximum wait time in milliseconds (default: 5 seconds).
+	 * @param delayMs initial delay time in milliseconds (default: 100ms).
+	 * @returns a promise that resolves when the file exists or rejects when the timeout is reached.
+	 */
+	protected async waitForFileExists(fileUri: Uri, maxWaitMs: number = 5_000, delayMs: number = 100): Promise<void> {
```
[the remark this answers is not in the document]
djelinek: hmmm this is weird.. it should cover only short time between long jbang task execution and fs file existence..

anyways nice catch, thanks!

djelinek:
> * given that we cannot create afile when no folde ris opened, we should either hide the "new file" action from the view or provide a message mentioning it. Currently ther eis avaliadation message on the second step about a "workspace folder undefined"

fair point, I have added notification which should instruct user what is a requirement

image

src/commands/AbstractNewCamelRouteCommand.ts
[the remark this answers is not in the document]
djelinek: seems root issue was a single missing `await`... we should really improve ESlint setup for vscode-kaoto project 😱

it should be fine now

djelinek:
for a sake of getting things done faster into main I would like to skip `extensionWeb` entrypoint (Web Extension) part for the moment.. at the moment the problem is Web Extensions are nto supporting `ShellExecution` API which is used for JBang tasks execution in VS Code. It could be potentially reworked into `CustomExecution` but it would cost much more time... WDYT? @apupier @lhein @lordrip 

for a context, below is what **Web Extensions** stands for:


### Web Extensions

Visual Studio Code can run as an editor in the browser. One example is the github.dev user interface reached by pressing . (the period key) when browsing a repository or Pull Request in GitHub. When VS Code is used in the Web, installed extensions are run in an extension host in the browser, called the 'web extension host'. An extension that can run in a web extension host is called a 'web extension'.

Web extensions share the same structure as regular extensions, but given the different runtime, don't run with the same code as extensions written for a Node.js runtime. Web extensions still have access to the full VS Code API, but no longer to the Node.js APIs and module loading. Instead, web extensions are restricted by the browser sandbox and therefore have [limitations](https://code.visualstudio.com/api/extension-guides/web-extensions#web-extension-main-file) compared to normal extensions.

The web extension runtime is supported on VS Code desktop too. If you decide to create your extension as a web extension, it will be supported on [VS Code for the Web](https://code.visualstudio.com/docs/editor/vscode-web) (including vscode.dev and github.dev) as well as on the desktop and in services like [GitHub Codespaces](https://code.visualstudio.com/docs/remote/codespaces).

lhein:
I am fine with ignoring it for the moment. Our main distribution way is as VS Code extension for now.

lhein (approved):
checked from a functional pov

lordrip:
@djelinek I think it makes sense, I just have a doubt now, considering that we don't have the same support level when running as a web extension, is it possible to simply remove the unsupported commands? Let's say creating new files using JBang is not supported because there's no shell to run it, would it be possible to avoid showing the command in the first place?

djelinek:
> @djelinek I think it makes sense, I just have a doubt now, considering that we don't have the same support level when running as a web extension, is it possible to simply remove the unsupported commands? Let's say creating new files using JBang is not supported because there's no shell to run it, would it be possible to avoid showing the command in the first place?

totally right question! that is something I am trying to implement now. Basically it is about to "hide" buttons when we are in web extension context.. but.. I am not sure how it will work and unfortunately it cannot be tetted locally it seems so now I am trying to test in vscode-dev web environment 🙁 ..

plan B?.. remove extension for web for a moment until we decide to fully support it..

djelinek:
my doubts are mainly about options I can see.. there are builtin context values in VS Code API which seems to be related

- `isWeb`	True when accessing the editor from the Web.
- `isVirtualWorkspace`

but I am afraid it will affect all web approaches which is also eg Red Hat Dev Spaces and so on..

djelinek:
After raising concerns with @apupier, we would suggest not conditioning desktop extension vs web extension for a moment. And leave new features for both options while primary needs to be tested and fully working on desktop. We can raise the issue to investigate the problem later.

This approach is the best way to avoid blocking the process of getting things done in the main branch. It seems to be a best effort solution at the moment.

cc @lhein @lordrip

djelinek:
UI tests are finally working! PR is ready for final review 🙂

src/extension/ExtensionContextHandler.ts
```diff
@@ -58,5 +61,19 @@ export class ExtensionContextHandler {
...
+        // register commands for new Camel files creation using YAML DSL - Routes, Kamelets, Pipes
+        this.context.subscriptions.push(vscode.commands.registerCommand(NewCamelRouteCommand.ID_COMMAND_CAMEL_ROUTE_YAML, async () => {
+            await new NewCamelRouteCommand('YAML').create();
+        }));
+        this.context.subscriptions.push(vscode.commands.registerCommand(NewCamelKameletCommand.ID_COMMAND_CAMEL_KAMELET_YAML, async () => {
+            await new NewCamelKameletCommand('YAML').create();
+        }));
+        this.context.subscriptions.push(vscode.commands.registerCommand(NewCamelPipeCommand.ID_COMMAND_CAMEL_PIPE_YAML, async () => {
+            await new NewCamelPipeCommand('YAML').create();
+        }));
```
apupier, line 77: for another iteration, provide telemtry https://issues.redhat.com/browse/FUSETOOLS2-2552

src/extension/ExtensionContextHandler.ts
```diff
@@ -58,5 +61,19 @@ export class ExtensionContextHandler {
         });
         this.context.subscriptions.push(integrationsTreeView);
         this.context.subscriptions.push(vscode.commands.registerCommand('kaoto.integrations.refresh', () => integrationsProvider.refresh()));
+        this.registerNewCamelYamlFiles();
```
apupier, line 64: picky: I would named it `registerNewCamelYamlFilesCommands` to avoid risk of potential confusion of creating some specific yaml files

apupier (approved):

PR #814, opened 2025-03-05
feat: add right-click options for each file item in Kaoto: Integrations view

djelinek:
Context Menu options
- delete file
- show source

Screenshot 2025-03-05 at 15 03 08


lhein (approved):
works as expected

PR #816, opened 2025-03-07
feat: add export inline button into Kaoto: Integrations view

djelinek:
https://github.com/user-attachments/assets/a0215d11-ada8-452b-88fd-42a1d3991411

### DIFF
--- a/it-tests/Z_IntegrationsViewNewProject.test.ts
+++ b/it-tests/Z_IntegrationsViewNewProject.test.ts
@@ -1,0 +1,150 @@
+/**
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { expect } from 'chai';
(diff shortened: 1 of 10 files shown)


lhein (approved):

lhein (approved):

it-tests/Z_IntegrationsViewNewProject.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+			// GAV selection
+			input = await InputBox.create(30_000);
+			await input.confirm();
+
+			// specify new project output dir
+			input = await InputBox.create(30_000);
+			await input.setText(PROJECT_OUTPUT_DIR);
+			await input.confirm();
+			await input.confirm(); // from some reason when using setText for a path pick input it needs to be confirmed twice (see https://github.com/redhat-developer/vscode-extension-tester/issues/1778)
```
apupier, line 105:
```suggestion
			await input.confirm(); // for some reason when using setText for a path pick input it needs to be confirmed twice (see https://github.com/redhat-developer/vscode-extension-tester/issues/1778)
```

resources/quarkus/launch.json
```diff
@@ -0,0 +1,18 @@
...
+            "name": "Run Camel Quarkus JVM application and attach Camel debugger",
+            "type": "apache.camel",
+            "request": "attach",
+            "preLaunchTask": "Start Camel application with Maven Quarkus Dev with camel.debug profile"
+        },
+        {
+            "name": "Run Camel native application and attach Camel debugger",
+            "type": "apache.camel",
+            "request": "attach",
+            "preLaunchTask": "Start Camel native application debug-ready"
```
apupier, line 14: does it make sense to provide the launch configuration with attach debugger if we push for not having the extension pack? (the launch configuration won't work without the VS Code debug adapter for Camel)
Or maybe we need to find a way to mention that the other extension is mandatory

djelinek: good point.. I will remove .vscode creation

apupier (approved):

PR #826, opened 2025-03-12
fix: improve regex for a GAV validation of new project creation process

djelinek:
from now it should be more compliant with common maven project conventions for GAV

groupId - The is expected to follow the Java package naming conventions
artifactId - The identifiers should only consist of lowercase letters, digits, and hyphens
version - The version is expected be compliant with Semantic Versioning 1.0.0

### DIFF
--- a/src/commands/NewCamelProjectCommand.ts
+++ b/src/commands/NewCamelProjectCommand.ts
@@ -79,7 +79,7 @@ export class NewCamelProjectCommand {
 	 * @param name
 	 * @returns string | undefined
 	 */
-	private validateGAV(name: string): string | undefined {
+	public validateGAV(name: string): string | undefined {
 		if (!name) {
 			return 'Please provide a GAV for the new project following groupId:artifactId:version pattern.';
 		}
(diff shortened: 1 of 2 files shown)


it-tests/Z_IntegrationsViewNewProject.test.ts
```diff
@@ -96,6 +96,7 @@ describe('Integrations View', function () {
 
 			// GAV selection
 			input = await InputBox.create(30_000);
+			await input.setText('com.acme:my-project:1.0-SNAPSHOT');
```
apupier, line 99: I think it is more important to continue testing with the default value here. And have a set of unit tests for the various validation cases?

src/commands/NewCamelProjectCommand.ts
```diff
@@ -95,22 +95,22 @@ export class NewCamelProjectCommand {
 			return 'The groupId cannot start with a .';
 		}
 		for (const groupIdSubPart of groupIdSplit) {
-			const regExpSearch = /^[a-z]\w*$/.exec(groupIdSubPart);
+			const regExpSearch = /^(?!\.)[a-z0-9]+(\.[a-z0-9]+)*$/.exec(groupIdSubPart);
 			if (regExpSearch === null || regExpSearch.length === 0) {
-				return `Invalid subpart of groupId: ${groupIdSubPart}} . It must follow groupId:artifactId:version pattern with groupId subpart separated by dot needs to follow this specific pattern: [a-zA-Z]\\w*`;
+				return `Invalid subpart of groupId: '${groupIdSubPart}'. The is expected to follow the Java package naming conventions.`;
```
apupier, line 100: "The is"? I guess some words are missing

apupier (changes requested):

apupier (approved):
"just" need to backport the change to the VS Code Language support extension also

PR #832, opened 2025-03-14
removed the test for a specific number of tabs

lhein:
This test will break with every change to the number of tabs in the editor. IMO it should be fully removed.


lhein:
closing in favor of the PR Dominik has already prepared on his side.

PR #833, opened 2025-03-14
fix(test): remove Kaoto UI tabs precise number check

djelinek:
to unblock build with a main kaoto branch


lhein (approved):

apupier (changes requested):
This sounds an important check that all expected tabs are available

djelinek:
> This sounds an important check that all expected tabs are available

the check is basically still in place. We are trying to switch to each tab - Design, Beans and DataMapper (except new About tab which I can add in another PR)

PR #834, opened 2025-03-17
feat: add Kubernetes Run button into Kaoto: Integrations view

djelinek:
image


lhein (approved):

README.md
```diff
@@ -41,6 +41,7 @@
 - Setting to provide custom set of catalog
 - Export single integration file into Maven project (Quarkus or Spring Boot)
 - Run single integration file locally
+- Deploy single integration file into Kubernetes cluster
```
apupier, line 44: I think we could add Kubernetes/OpenShift to make it clearer that it works with OpenShift (which is even by default)

apupier (commented):
It might be worthy to add a little chapter to explain how to deploy on Kubernetes with minikube given that it is a lot more complicated

djelinek:
> It might be worthy to add a little chapter to explain how to deploy on Kubernetes with minikube given that it is a lot more complicated

i will mention an upstream reference [1] and it should bring the value we need. Then just need to double check it is up to date in upstream

[1] https://camel.apache.org/manual/camel-jbang-kubernetes.html#_minikube_deployment_tips_and_troubleshooting

apupier (approved):

PR #837, opened 2025-03-19
fix: respect parent folder of the running integration when looking for existing xsl files

djelinek:
we need to search xsl files in current folder of running integration file instead of always searching whole workspace

### DIFF
--- a/src/helpers/CamelJBang.ts
+++ b/src/helpers/CamelJBang.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ShellExecution, ShellExecutionOptions, Uri, workspace, WorkspaceFolder } from 'vscode';
+import { ShellExecution, ShellExecutionOptions, workspace } from 'vscode';
 import { arePathsEqual } from './helpers';
 import { dirname } from 'path';
 import { globSync } from 'glob';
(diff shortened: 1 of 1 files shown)


lhein (approved):

src/helpers/CamelJBang.ts
```diff
@@ -147,8 +147,7 @@ export class CamelJBang {
 	 * it is caused by null glob option disabled by default for ZSH shell
 	 */
 	private handleMissingXslFiles(filePath: string, runArgs: string[]): string[] {
-		const currentFileWorkspace = workspace.getWorkspaceFolder(Uri.file(filePath)) as WorkspaceFolder;
-		const xsls = globSync(`${currentFileWorkspace.uri.path}/**/*.xsl`).length > 0;
+		const xsls = globSync(`${dirname(filePath)}/*.xsl`).length > 0;
```
apupier, line 150: is it expected to no more search in subfolders?

apupier (commented):
Is similar change needed for VS Code Language Support?

src/helpers/CamelJBang.ts
djelinek: yeah I see your point.. but the thing is Kaoto DataMapper is generating XSL files in same path as the route where you added datamapper step

it means when we are scanning all subfolders it will affect running route without datammaper too, see potential structure below

option A:
```
- workspace
  - folder A
    - routeA.camel.yaml // route without datamapper step
  - folder B
    - routeB.camel.yaml // route with datamapper step
    - datamapper.xsl
```

option B:
```
- workspace
  - folder A
    - routeA.camel.yaml // route without datamapper step
    - folder B
      - routeB.camel.yaml // route with datamapper step
      - datamapper.xsl
```

if you will try to run single file on `routeA.camel.yaml` it will fail, because in both cases if you will scan also subfolders for a workspace/parent folder the xsl file is present

djelinek:
> Is similar change needed for VS Code Language Support?

yeah this issue is same for current Debug Adapter extension implementation - see https://github.com/camel-tooling/camel-dap-client-vscode/blob/main/src/task/CamelJBangTaskProvider.ts#L278

lordrip (approved):

src/helpers/CamelJBang.ts
apupier: ah right, currently fo rthe run we provide single-level config too https://github.com/KaotoIO/vscode-kaoto/blob/76ae6ae0ffa3aa263b8d71e295198f0075dd8da6/package.json#L416

(which means that if peole put them in a subfolder it won't work anymore but they had to change the defaults)

apupier (approved):

PR #842, opened 2025-03-24
FUSETOOLS2-253 - enable test for export on MacOS

apupier:


djelinek (approved):
nice! that is something I didn't noticed before. Thanks!

djelinek:
interesting thing.. for me locally this change is not working because I am having `Select` button on my mac visible

apupier:
maybe depends on Mac version?

PR #849, opened 2025-03-26
fix: web extension failed to load in a browser

djelinek:
Screenshot 2025-03-26 at 12 49 10

### DIFF
--- a/src/helpers/CamelJBang.ts
+++ b/src/helpers/CamelJBang.ts
@@ -13,10 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ShellExecution, ShellExecutionOptions, workspace } from 'vscode';
+import { RelativePattern, ShellExecution, ShellExecutionOptions, Uri, workspace } from 'vscode';
 import { arePathsEqual } from './helpers';
 import { dirname } from 'path';
-import { globSync } from 'glob';
 
 export enum RouteOperation {
 	start = 'start',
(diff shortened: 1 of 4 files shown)


src/helpers/CamelJBang.ts
```diff
@@ -146,7 +139,9 @@ export class CamelJBang {
 	 * Mainly in ZSH shell there is problem when Camel JBang is executed with non existing files added using '*.xsl' file pattern
 	 * it is caused by null glob option disabled by default for ZSH shell
 	 */
-	private handleMissingXslFiles(filePath: string, runArgs: string[]): string[] {
+	private async handleMissingXslFiles(filePath: string, runArgs: string[]): Promise<string[]> {
+		const globSync = (await import('glob')).globSync; // glob needs to be imported dynamically to avoid issues when extension is bundled for Web
```
apupier, line 143: ouch this smells problems with edge cases

djelinek: which edge cases?

this all mess is just about somehow divide web and desktop extension bundles.. other way is to strictly divide codebase which is touching extension points - extension.ts and extensionWeb.ts

which is possible in term of full project refactor

this dynamic loading happens when code is really called which is at the moment happening only for desktop (extension.ts)

apupier (approved):
works but I do not understand well why

src/helpers/CamelJBang.ts
apupier: by experience, using dynamic import inlined in the code is often not working in some conditions

lhein: Can you remind me why we need that web extension thingy at all?

djelinek: if I understood correctly the main reason is Github Codespaces or vscode.dev in a browser

main usecase for GH Codespace would be to open it by press ´.´ on some camel gh repository which will opens it automatically using codespaces and then when you install Kaoto extension there you should be able to design/see routes

djelinek: how often this usage is expected or how useful it is I do not know..

djelinek: alright.. I have removed `glob` dependency to avoid need of using dynamic loading 🙂 now we are relying on a vscode `workspace` API which should be much safer

lhein (approved):

apupier (approved):

apupier (changes requested):
ah await need to verify if we can remove glob from package.json dependencies

apupier (approved):
glob is not listed in package.json (so guess it was working thanks to implicit transitive dependencies before)

PR #851, opened 2025-03-26
test: fix unstable export project UI test mainly on macOS

djelinek:
in some cases there is test still failing because of another button, depending probably on platform, localization and others to ensure well formatted path inside input box


djelinek:
follow up on #842

lhein (approved):

apupier (approved):

PR #854, opened 2025-03-28
feat: add Stop and Logs action buttons into Kaoto: Deployments view

djelinek:
adding `Stop` and `Follow Logs` buttons

image

REPORTED upstream issue - https://issues.apache.org/jira/browse/CAMEL-21921

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
     - create new Camel Route using XML DSL
     - list XML files with route elements
 - Add `Deployments` view into Kaoto view container which displays all local running integrations
+- Add `Stop` and `Follow Logs` action buttons into Deployments view
 
 # 2.4.0
 
--- a/it-tests/views/06_DeploymentsViewLogsAndStop.test.ts
+++ b/it-tests/views/06_DeploymentsViewLogsAndStop.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Copyright 2025 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { expect } from 'chai';
+import { join } from 'path';
+import {
+	ActivityBar,
+	after,
+	before,
+	BottomBarPanel,
+	EditorView,
+	SideBarView,
+	TreeItem,
+	ViewControl,
+	ViewSection,
+	VSBrowser,
+	WebDriver,
+} from 'vscode-extension-tester';
+import { getTreeItem, killTerminal, waitUntilTerminalHasText } from '../Util';
+
+describe('Deployments View', function () {
+	this.timeout(600_000); // 10 minutes
+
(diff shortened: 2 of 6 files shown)


apupier:
fixes https://issues.redhat.com/browse/FUSETOOLS2-2558 and https://issues.redhat.com/browse/FUSETOOLS2-2560

apupier (changes requested):
found one issue:
* run the integration
* stop the integration
* run again the same integration
* use show log
--> 2 terminal are named the same because the one from the previous run is still here. it can point to this one instead of the running instance

djelinek:
> found one issue:
> 
> * run the integration
> * stop the integration
> * run again the same integration
> * use show log
>   --> 2 terminal are named the same because the one from the previous run is still here. it can point to this one instead of the running instance

I cannot reproduce 🙁 for me the stopped terminal has not same name as next executed
image

could you provide me some reproducer please?

djelinek:
> > found one issue:
> > 
> > * run the integration
> > * stop the integration
> > * run again the same integration
> > * use show log
> >   --> 2 terminal are named the same because the one from the previous run is still here. it can point to this one instead of the running instance
> 
> I cannot reproduce 🙁 for me the stopped terminal has not same name as next executed image
> 
> could you provide me some reproducer please?

@apupier I have improved lookup condition and it should now consider also if terminal to be activated is active.. could you please verify if it is working better for you now?

apupier:
> > found one issue:
> > 
> > * run the integration
> > * stop the integration
> > * run again the same integration
> > * use show log
> >   --> 2 terminal are named the same because the one from the previous run is still here. it can point to this one instead of the running instance
> 
> I cannot reproduce 🙁 for me the stopped terminal has not same name as next executed image
> 
> could you provide me some reproducer please?

* run the integration with the inline button in Integrations view
* stop the integration using the stop inlined button in Deployments view
* run again the same integration with the inline button in Integrations view
* use show log inlined button in Deployment view
--> 2 terminal are named the same because the one from the previous run is still here. it can point to this one instead of the running instance


[wrong terminal selected.webm](https://github.com/user-attachments/assets/2a1935be-6d4e-4b0a-9851-68340c7b1cf3)

apupier:
> @apupier I have improved lookup condition and it should now consider also if terminal to be focused is active.. could you please verify if it is working better for you now?

still same issue

djelinek:
> reproduced the issue. Just follow the steps Aurelien shared:
> 
> 1. Launch the integration
> 2. Terminate the integration - but do not close the terminal
> 3. Launch the same integration
> 4. Use the "Follow Log" button
> 5. see that it activates to terminal tab that has the terminated integration

alright, I have reproduced locally with an ugly workaround (probably OS related glitchy behavior). Anyways.. unfortunately there is no way atm to fix it in current implementation.

however it will be fixed after #855 because the terminal name will be more unique by port in name of it `Running - test.camel.yaml:56595`. The port is generated randomly for each execution so it should avoid this potential issue.

lhein:
Another approach (or maybe even as addition) could be to close the terminal once you terminated the integration.

djelinek:
> Another approach (or maybe even as addition) could be to close the terminal once you terminated the integration.

yeah that is possible but in that way you loss the logs.. so in case you just wanted to stop and read or copy from logs it won't be possible anymore

djelinek:
> > @apupier I have improved lookup condition and it should now consider also if terminal to be focused is active.. could you please verify if it is working better for you now?
> 
> still same issue

ready for a final review, the focus terminal issue should be fixed now

lhein:
Works perfectly

apupier (approved):
- while testing it, i realized the impact of https://issues.apache.org/jira/browse/CAMEL-21921 is very important, I think we need to put a note as known issues in readme or documentation
- The fact that there is no difference visible between terminal with stopped integrations and running one is not very convenient, I reported https://issues.redhat.com/browse/FUSETOOLS2-2581

PR #868, opened 2025-04-03
chore(deps): remove unused monaco-* dependencies

djelinek:
those monaco-* dependencies are used only in Kaoto UI for source code editor in web


lordrip (approved):
_Don't you love the smell of removing dependencies in the morning?_ :smiley:

apupier (changes requested):
awesome.
Can be removed from the dependabot config too

apupier (approved):

PR #879, opened 2025-04-08
Fix deploy on OpenShift by default on restricted environments

apupier:
Camel 4.11 has introduced `--disable-auto` parameter without checking if the cluster-type has been already set or not.
The automatic detection does not work if connected user cannot list resource "clusterversions" in API group "config.openshift.io" at the cluster scope.

Other better ways that could be done later on:
* remove automatically the --disable-auto when detecting that the version is prior to 4.11
* forget about 4.11 and improve handling these parameters in Camel JBang


apupier:
successful build on Jenkins CI https://jenkins-csb-fusetools-qe-master.dno.corp.redhat.com/view/VS%20Code%20-%20release/job/vscode/job/eng/job/vscode-kaoto-release/212/

djelinek (approved):

PR #884, opened 2025-04-10
Adapt UI test to Kaoto main

apupier:
the data-testid for Datamapper step has been modified, it now contains also the path in the structure

### DIFF
--- a/it-tests/BasicFlow.test.ts
+++ b/it-tests/BasicFlow.test.ts
@@ -5,6 +5,8 @@ import { checkEmptyCanvasLoaded, checkTopologyLoaded, openAndSwitchToKaotoFrame,
 import { waitUntil } from 'async-wait-until';
 import * as fs from 'fs-extra';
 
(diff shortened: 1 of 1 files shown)


it-tests/BasicFlow.test.ts
```diff
@@ -5,6 +5,8 @@ import { checkEmptyCanvasLoaded, checkTopologyLoaded, openAndSwitchToKaotoFrame,
 import { waitUntil } from 'async-wait-until';
 import * as fs from 'fs-extra';
 
+const DATA_TEST_ID_DATAMAPPERSTEP_2_5 = 'custom-node__route.from.steps.0.step:kaoto-datamapper';
```
lhein, line 8: why marking it with a version 2_5?

apupier: to know that it was introduced with 2_5, so that we can potentially remove and clean up previous versions more easily when a kaoto version is released

PR #904, opened 2025-05-06
chore: in a dev mode disable other extensions for an Extension Host vscode instance

djelinek:
for those interested just add into your .bashrc/.zshrc line

`export VSCODE_DISABLE_EXT="--disable-extensions"`

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
       "type": "extensionHost",
       "request": "launch",
       "runtimeExecutable": "${execPath}",
-      "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
+      "args": ["${env:VSCODE_DISABLE_EXT}", "--extensionDevelopmentPath=${workspaceRoot}"],
       "outFiles": ["${workspaceRoot}/dist/**/*.js"],
       "preLaunchTask": "npm: watch",
       "env": {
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -6,7 +6,7 @@
       "type": "extensionHost",
       "request": "launch",
       "runtimeExecutable": "${execPath}",
-      "args": ["--extensionDevelopmentPath=${workspaceRoot}"],
+      "args": ["--disable-extensions", "--extensionDevelopmentPath=${workspaceRoot}"],
```
apupier, line 9: By default, I think it is is better to keep all extensions, it is closer to real-world use case

djelinek: I mean it depends.. if your vscode instance has more extensions you work with then having all those also in extensions host enabled by default is not efficient.. but as you prefer.. I am doing this manually anyways because I do not want to have two instances at same time with all the stuff

lordrip: I like the idea of having a VS Code launch without any other extensions, and probably both scenarios has their use.

Couldn't we do something like this?
```
"args": ["--disable-extensions=${VSCODE_DISABLE_EXT}", "--extensionDevelopmentPath=${workspaceRoot}"],
```

While `VSCODE_DISABLE_EXT` is an environment variable that we could set to `true` so it works for us, and you simply wouldn't have?

djelinek: seems I've found some working approach
`"args": ["${env:VSCODE_DISABLE_EXT}", "--extensionDevelopmentPath=${workspaceRoot}"],`

with locally `export VSCODE_DISABLE_EXT="--disable-extensions"`

lordrip: mmm, in that case, I guess it would be more appropriate to call it `VSCODE_LAUNCH_ARGS` or similar

djelinek: not necessarily.. proposed env variable contains only one arg related to disable extensions

PR #930, opened 2025-05-29
fix(Theme): Configure color theme

lordrip:
fix: https://github.com/KaotoIO/kaoto/issues/2305

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,6 +500,16 @@
               "onHover",
               "onSelection"
             ]
+          },
+          "kaoto.colorScheme": {
+            "type": "string",
+            "markdownDescription": "Choose the color scheme for the UI. Can be either `auto`, `light`, or `dark`",
+            "default": "ligth",
+            "enum": [
+              "auto",
+              "light",
+              "dark"
+            ]
           }
         }
       },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -500,6 +500,16 @@
               "onHover",
               "onSelection"
             ]
+          },
+          "kaoto.colorScheme": {
+            "type": "string",
+            "markdownDescription": "Choose the color scheme for the UI. Can be either `auto`, `light`, or `dark`",
+            "default": "light",
```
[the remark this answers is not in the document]
lordrip: I wanted to use `auto` as the default, but as a couple of users reported that they liked how it was originally, I picked `light` instead.

package.json
```diff
@@ -500,6 +500,16 @@
               "onHover",
               "onSelection"
             ]
+          },
+          "kaoto.colorScheme": {
```
djelinek, line 504: picky not:

- `Color Theme` would be more precise if we would like to follow VS Code naming pattern

package.json
```diff
@@ -500,6 +500,16 @@
...
+            "type": "string",
+            "markdownDescription": "Choose the color scheme for the UI. Can be either `auto`, `light`, or `dark`",
```
apupier, line 506: I think we should mention:
* what means auto (picking the category them of the VS Code instance)
* that the editor must be closed and reopen to apply the change

apupier (changes requested):
- worth adding a note in the changelog

package.json
lordrip: I used the same config setting that we're using in Kaoto UI. We could change the name in VS Code and make the adaptation where the VS Code MA API is created. We might be deviating from the other settings where we use the exact same name, but maybe is worth the small disparity to offer a more VS Code-like experience :+1: 

I'll change it

lordrip: done

package.json
lordrip: done

apupier: if I'm not mistaken, it is still not mentioned int he visible part to the users, in the description of the setting

djelinek (approved):

apupier (approved):

PR #945, opened 2025-06-13
Provide conditional wait to check for notifications to stabilize tests

apupier:


apupier:
given that the screenshot is showing the XML notification remaining, that the test for xml has passed and that for yaml, it fails to find it. I'm suspecting that when new notifications are arriving they can mess up the other existing notifications and so the click on the button is done on a wrong notification tile

djelinek (approved):

PR #957, opened 2025-06-27
Provide function to retrieve Runtime version information on Maven projects

apupier:
[KTO-494](https://issues.redhat.com/browse/KTO-494)


src/helpers/CamelJBang.ts
```diff
@@ -123,6 +126,42 @@ export class CamelJBang {
 		return new ShellExecution(this.jbang, [...this.defaultJbangArgs, 'cmd', `${operation}-route`, integration, `--id=${routeId}`]);
 	}
 
+	public async getRuntimeInfoFromMavenContext(integrationFilePath: string): Promise<string | undefined> {
+		const folderOfpomXml = this.findFolderOfPomXml(integrationFilePath);
+		if (folderOfpomXml !== undefined) {
+			try {
+				let camelJbangVersionToUse: string;
+				if (compareVersions(this.camelJBangVersion, '4.12')) {
```
lordrip, line 134: JBang 4.12 doesn't consider multiple `pom.xml` files, would it be better to use `4.13` instead?

src/helpers/CamelJBang.ts
```diff
@@ -123,6 +126,42 @@ export class CamelJBang {
 		return new ShellExecution(this.jbang, [...this.defaultJbangArgs, 'cmd', `${operation}-route`, integration, `--id=${routeId}`]);
 	}
 
+	public async getRuntimeInfoFromMavenContext(integrationFilePath: string): Promise<string | undefined> {
```
apupier, line 129: currently providing a string which is the full json string returned by Camel JBang. Potentially possible to provide a specific object. It will be more Type-safed and easier to manipulate. In this case, not sure where this Type should be defined.

src/helpers/CamelJBang.ts
lordrip: Didn't the keys change between versions as well?

lordrip: I tried with 4.12 and 4.13-SNAPSHOT, and the main difference is that for 4.13 we need to specify the `pom.xml` file of the project like:
```
camel dependency runtime ~/folder/pom.xml
```

Structure-wise, these are the differences; the main keys remain unchanged.

#### 4.12
```json
{
  "runtime": "camel-quarkus",
  "camelVersion": "4.11.0",
  "camelQuarkusVersion": "3.22.0",
  "quarkusVersion": "3.22.2"
}
```

#### 4.13
```json
{
  "runtime": "quarkus",
  "camelVersion": "4.11.0",
  "camelQuarkusVersion": "3.22.0",
  "quarkusVersion": "3.22.2",
  "quarkusBomGroupId": "io.quarkus.platform",
  "quarkusBomArtifactId": "quarkus-bom",
  "camelQuarkusBomGroupId": "io.quarkus.platform",
  "camelQuarkusBomArtifactId": "quarkus-camel-bom"
}
```

apupier:
```
is that for 4.13 we need to specify the pom.xml 
```
argh it means we have an API break on the first new release with the feature :'-(

lordrip: The 4.12 version is barely usable, I think we should skip it and focus on 4.13 onwards

src/helpers/CamelJBang.ts
lordrip: Yes, you're right, Kaoto should define the type, and then the host application should fulfil it.

PR #973, opened 2025-07-10
feat: implement getSuggestions API method to return available OS env variables

djelinek:
relates to https://github.com/KaotoIO/kaoto/pull/2354

### DIFF
--- a/src/helpers/SuggestionRegistry.ts
+++ b/src/helpers/SuggestionRegistry.ts
@@ -1,0 +1,85 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License", destination); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type Suggestion = {
+	value: string;
+	description?: string;
+	group?: string;
+};
+
+export type SuggestionContext = {
+	propertyName: string;
+	inputValue: string | number;
+	cursorPosition?: number | null;
+};
+
+export type SuggestionProviderFunction = (word: string, context: SuggestionContext) => Suggestion[] | Promise<Suggestion[]>;
+
+const suggestionRegistry = new Map<string, SuggestionProviderFunction>();
+
+/**
+ * Registers a new suggestion provider for a given topic
+ */
+export function registerSuggestionProvider(topic: string, providerFn: SuggestionProviderFunction): void {
+	suggestionRegistry.set(topic, providerFn);
+}
+
+/**
+ * Retrieves suggestions from the provider registered for the given topic
+ * @param topic The topic for which suggestions are being requested (e.g., "env", "properties", "kubernetes", "beans", etc.)
+ * @param word The current word or input value for which suggestions are being requested
+ * @param context Additional context for the suggestions, such as the property name and current input value.
+ * @returns A promise that resolves to an array of suggestions, each containing a value, optional description, and optional group.
+ */
+export async function getSuggestions(topic: string, word: string, context: SuggestionContext): Promise<Suggestion[]> {
+	const suggestionProvider = suggestionRegistry.get(topic);
+	if (!suggestionProvider) {
+		return [];
+	}
+	const result = suggestionProvider(word ?? '', context);
+	return Promise.resolve(result);
+}
+
+const provideEnvSuggestions: SuggestionProviderFunction = (word, _context) => {
+	const lowerWord = word.toLowerCase();
+	return Object.keys(process.env)
+		.map((envVar) => {
+			if (envVar.startsWith(word)) {
+				return { envVar, rank: 0 }; // Exact case-sensitive prefix match
+			}
+			if (envVar.toLowerCase().startsWith(lowerWord)) {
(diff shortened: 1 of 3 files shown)


src/helpers/SuggestionRegistry.ts
```diff
@@ -0,0 +1,85 @@
...
+export type SuggestionProviderFunction = (word: string, context: SuggestionContext) => Suggestion[] | Promise<Suggestion[]>;
+
+const suggestionRegistry = new Map<string, SuggestionProviderFunction>();
+
+/**
+ * Registers a new suggestion provider for a given topic
+ */
+export function registerSuggestionProvider(topic: string, providerFn: SuggestionProviderFunction): void {
+	suggestionRegistry.set(topic, providerFn);
+}
```
lordrip, line 39: [friendly-comment] Not strictly necessary right now, but if in the future we need to clear this registry, we can't.

I can see 2 scenarios where we might want to unregister a provider, one when there's a disconnection from one of the sources, and maybe for testing purposes.

If we reach one of those scenarios, we could integrate the `suggestionRegistry` and the `registrySuggestionProvider` function into a class so we could offer an `unregister` method as well.

src/test/webview/suggestionProviders.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+		setup(() => {
+			originalEnv = { ...process.env };
+
+			// define test environment variables
+			process.env.TEST_ALPHA = 'A';
+			process.env.TEST_BETA = 'B';
+			process.env.TESTCASE = 'C';
+			process.env.Test_Mixed = 'mixed';
+			process.env.PATH = '/usr/local/bin';
```
lordrip, line 32: Do we have a test about writing `Te` and then getting `Tst_Mixed` first, and then the others?

lordrip (approved):
Looks good to me, once this is merged, we can merge Kaoto side

apupier:
What is the reason of passing through VS Code extension to access environment variable?
Given that they are accessible through node.js and so Kaoto directly https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs

src/test/webview/suggestionProviders.test.ts
djelinek: it was kinda covered, but I have improved second test to cover that more specifically - https://github.com/KaotoIO/vscode-kaoto/pull/973/files#diff-6d0299b63280917e832dd28caba420b15b5aef063be49b170fd12c8a96337f97R48

lordrip:
> What is the reason of passing through VS Code extension to access environment variable? Given that they are accessible through node.js and so Kaoto directly https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs

@apupier: To offer the environment variables as completions for the configuration panel.

lordrip:
We have what we need for the UI part, I'll merge this PR at the end of the day unless you have a blocker @apupier :+1:

apupier:
> > What is the reason of passing through VS Code extension to access environment variable? Given that they are accessible through node.js and so Kaoto directly https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs
> 
> @apupier: To offer the environment variables as completions for the configuration panel.

My question is why going through VS Code extension to retrieve the environment variable? It sounds it can be retrieved directly by Kaoto using the pointed API. Or is it that these environment variables are not visible in the Node.js context of Kaoto?

lordrip:
> ugh

No, unfortunately, we can't read process.env from Kaoto UI because it runs in the browser, not as a host process; that's why we're using the VS Code host.

src/helpers/SuggestionRegistry.ts
```diff
@@ -0,0 +1,85 @@
...
+	group?: string;
+};
+
+export type SuggestionContext = {
+	propertyName: string;
+	inputValue: string | number;
+	cursorPosition?: number | null;
+};
+
+export type SuggestionProviderFunction = (word: string, context: SuggestionContext) => Suggestion[] | Promise<Suggestion[]>;
```
apupier, line 30: Shouldn't these types declared on kaoto side? Otherwise, Kaoto won't be able to call them without introducing a cyclic dependency or redeclaring the types (not even sure it would work in this case)

src/helpers/SuggestionRegistry.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Retrieves suggestions from the provider registered for the given topic
+ * @param topic The topic for which suggestions are being requested (e.g., "env", "properties", "kubernetes", "beans", etc.)
+ * @param word The current word or input value for which suggestions are being requested
+ * @param context Additional context for the suggestions, such as the property name and current input value.
+ * @returns A promise that resolves to an array of suggestions, each containing a value, optional description, and optional group.
+ */
+export async function getSuggestions(topic: string, word: string, context: SuggestionContext): Promise<Suggestion[]> {
```
apupier, line 48: What is the interest of passing a topic as parameter?

If we provide a specific method for each of the topics, we won't need to lookup in a registry and directly call the correct suggestion provider.

lordrip: We could have done that, but that means we're gonna need a never-ending API to support each piece of information we want. We decided to use a topic instead to have a single method and avoid the tight coupling between VS Code Kaoto and Kaoto

src/webview/VSCodeKaotoEditorChannelApi.ts
```diff
@@ -14,6 +14,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { KaotoOutputChannel } from '../extension/KaotoOutputChannel';
 import { findClasspathRoot } from '../helpers/ClasspathRootFinder';
+import { getSuggestions as getSuggestionsFromRegistry } from '../helpers/SuggestionRegistry';
```
apupier, line 17: maybe better to keep the same method name?

src/helpers/SuggestionRegistry.ts
apupier: (and from the VS Code Kaoto side, I think that we can consider that the form library is part of Kaoto)

apupier (approved):

lordrip:
Thank you all :muscle:

PR #983, opened 2025-07-15
chore: add workspace setup for extensions

djelinek:


djelinek:
> mmm, can we add this to Kaoto as well? :)

I suppose yes.. in case you are using those extensions - mainly prettier, eslint and spell checker could be interesting for Kaoto

PR #986, opened 2025-07-18
Use newer online Catalog repository example

apupier:
the previous one was deprecated.

Note that the previous one is not compatible with current main version of Kaoto, see https://github.com/KaotoIO/kaoto/issues/2383


apupier:
fixing one problem with Kaoto main branch. It remains 15 tests failing

djelinek (approved):

PR #987, opened 2025-07-18
Fix compatibility tests with main branch of Kaoto which has renamed some data-testid in 2.7 from *-field-* to *-fx-*

apupier:


apupier:
the 1 remaining failing test on main branch of Kaoto is handled with this PR https://github.com/KaotoIO/vscode-kaoto/pull/986

PR #995, opened 2025-07-29
feat: add suggestions provider for application.properties files

djelinek:


src/helpers/ApplicationPropertiesFinder.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export async function findAllApplicationPropertiesFiles(startUri: vscode.Uri): Promise<vscode.Uri[]> {
+	let currentDir = path.dirname(startUri.fsPath);
+	const workspaceFolder = vscode.workspace.getWorkspaceFolder(startUri)?.uri.fsPath;
+
+	while (workspaceFolder && currentDir.startsWith(workspaceFolder)) {
+		const entries = await vscode.workspace.fs.readDirectory(vscode.Uri.file(currentDir));
+		const matchingFiles = entries
+			.filter(([name, type]) => {
+				return type === vscode.FileType.File && /^application(-.+)?\.properties$/.test(name);
```
lordrip, line 28: Creating RegExp objects is expensive in JavaScript, and considering we need to run this function several times when working with the configuration form, could we explore a slightly different mechanism to filter the files?

I'm wondering if this check could work:
```typescript
return type === vscode.FileType.File && name.startsWith('application') && name.endsWith('.properties');
```

Alternatively, we could create the RegExp once, outside of the function, and then use it and reset it immediately after:
```typescript
// Defined outside the function
const APP_PROPERTIES_REGEXP = /^application(-.+)?\.properties$/;

// Inside of the function
APP_PROPERTIES_REGEXP.lastIndex = 0;
return type === vscode.FileType.File && APP_PROPERTIES_REGEXP.test(name);
```

src/helpers/ApplicationPropertiesFinder.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export async function parseMultipleApplicationPropertiesFiles(fileUris: vscode.Uri[]): Promise<Suggestion[]> {
+	const suggestions: Suggestion[] = [];
+
+	for (const uri of fileUris) {
+		const bytes = await vscode.workspace.fs.readFile(uri);
+		const content = Buffer.from(bytes).toString('utf-8');
+		const fileName = path.basename(uri.fsPath);
+
+		for (const line of content.split('\n')) {
```
lordrip, line 54: Curious question: Would this work when working in Windows?

src/helpers/ApplicationPropertiesFinder.ts
```diff
@@ -0,0 +1,76 @@
...
+			const trimmed = line.trim();
+			if (!trimmed || trimmed.startsWith('#')) {
+				// skip comment and empty lines
+				continue;
+			}
+
+			const equalIndex = trimmed.indexOf('=');
+			if (equalIndex !== -1) {
+				const key = trimmed.slice(0, equalIndex).trim();
+				const valuePart = trimmed.slice(equalIndex + 1).trim();
```
lordrip, line 64: To be defined at a later stage: do we want to suggest empty properties or not?

I don't have a concrete answer at the moment, but I think this approach is fine, suggest what's defined and maybe populate it at a later stage

lordrip (commented):
It looks good 🙆‍♂️ , just a missing touches around the RegExp and we're good to go and wire it up at Kaoto UI side

src/helpers/ApplicationPropertiesFinder.ts
djelinek: problem for tests on windows was not related in the end.. thew thing was vscode-test cli cannot open (me and AI didn't find out..) workspace folder with spaces and special chars as we are using so I have used a bit workaround using vscode workspace definition and now it is working as expected

lordrip (approved):

PR #1022, opened 2025-08-21
Enable drag-and-drop by default in Kaoto VSCode

shivamG640:
Enable drag-and-drop by default in Kaoto VSCode

### DIFF
--- a/package.json
+++ b/package.json
@@ -519,7 +519,7 @@
         "properties": {
           "kaoto.enableDragAndDrop": {
             "type": "boolean",
-            "default": false,
+            "default": true,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -519,7 +519,7 @@
         "properties": {
           "kaoto.enableDragAndDrop": {
             "type": "boolean",
-            "default": false,
+            "default": true,
             "markdownDescription": "Control whether to enable drag and drop feature. It requires to reopen the Kaoto editors to be effective.",
             "tags": [
               "experimental"
```
djelinek, line 525: quick question: are we still in experimental phase of feature when enabling DnD by default? 🤔

PR #1032, opened 2025-09-02
fix: Camel JBang settings title in new vscode version is displayed wrong

djelinek:
BEFORE:
Screenshot 2025-09-02 at 13 07 26

AFTER:
Screenshot 2025-09-02 at 13 07 30

unfortunately there is no way to have it as before "JBang", after changes in latest vscode versions it displays this weird space 😞


apupier (changes requested):
a breaking change must be added to the changelog because users which had specific values for the settings won't be used anymore

lordrip (approved):

PR #1041, opened 2025-09-04
fix: bypass undo/redo shortcuts to allow disable them in Kaoto editor

djelinek:


lhein (approved):
This will be a temporary workaround until we have a fully working solution in place.

PR #1069, opened 2025-09-10
Use RuntimeMavenInformation type defined in Kaoto

### LINKED ISSUE #977: Use RuntimeMavenInformation type from Kaoto
The RuntimeMavenInformation type has been declared on VS Code Kaoto side first to avoid having phase with broken build. When it will be made available in Kaoto, use the oen from Kaoto

apupier:
instead of duplicating and redclaring it in VS Code Kaoto

fixes #977


lordrip (changes requested):
I think we should update the following instead:

https://github.com/KaotoIO/vscode-kaoto/blob/6af64f481535dd5e429b8daf579e2e3d66cefb8b/src/webview/VSCodeKaotoEditorChannelApi.ts#L206-L234

lordrip (approved):
Awesome 👍

PR #1070, opened 2025-09-10
Use Suggestion* types defined in Kaoto forms

### LINKED ISSUE #975: Use Suggestion API return type
The @kaoto/forms library exports a Suggestion return type; we should forward it through @kaoto/kaoto to VS Code to avoid duplicating the type.

After merging this: https://github.com/KaotoIO/kaoto/pull/2354 and releasing, we could import the `Suggestions`-related types from `@kaoto/kaoto`

apupier:
instead of duplicating and redeclaring it in VS Code Kaoto

fixes #975

### DIFF
--- a/src/helpers/ApplicationPropertiesFinder.ts
+++ b/src/helpers/ApplicationPropertiesFinder.ts
@@ -15,7 +15,7 @@
  */
 import vscode from 'vscode';
 import path from 'path';
-import { Suggestion } from './SuggestionRegistry';
+import { Suggestion } from '@kaoto/kaoto';
 
 export async function findAllApplicationPropertiesFiles(startUri: vscode.Uri): Promise<vscode.Uri[]> {
 	let currentDir = path.dirname(startUri.fsPath);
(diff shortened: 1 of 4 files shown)


djelinek:
the idea of mentioned issue was to bring those types from @kaoto/kaoto and not to rely on @kaoto/forms at the moment if I understood well when we were initially implementing that API provider. 

This PR aproach was in initial kickoff rejected if I remember correctly

apupier:
the most important part is to avoid redeclaring the same type in VS Code Kaoto side. Given that @kaoto/kaoto is re-exporting the type from @kaoto/forms for its own namespace, seems better to use @kaoto/kaoto than @kaoto/forms

djelinek (approved):

lordrip (changes requested):
We should update:

https://github.com/KaotoIO/vscode-kaoto/blob/6af64f481535dd5e429b8daf579e2e3d66cefb8b/src/webview/VSCodeKaotoEditorChannelApi.ts#L198-L204

lordrip (approved):
Thanks for updating the API👍

Note for the future: if we need to change the internal VS Code API, we can do so without affecting the contract between Kaoto UI and VS Code Kaoto 👍 , so we're not blocked internally.

PR #1078, opened 2025-09-16
feat(KIE): Override KIE KogitoEditorEnvelope

### LINKED ISSUE #1074: [KIE] Build Kaoto UI as a separate app for VS Code
Currently, the Kaoto UI is built with Webpack, and it uses the React bindings from KIE.

This is problematic as the Kaoto UI uses React 18, but KIE only supports React 17, so it forces the Kaoto UI to behave as if it were in React 17, causing some differences between VS Code and the web deployment.

The goal of this ticket is to build the Kaoto UI separately and use the generic KIE bindings so we don't need to downgrade React at runtime.

lordrip:
### Context
Currently, the `kie-tools` repository uses `React 17` which has a particular way to be bootstrapped, referenced now as `Legacy Root API`.

Kaoto UI uses `React 18` which uses a different bootstrap API called `New Root API`, and while it's backward compatible, the runtime behavior is the one from `React 17`.

This causes many subtles differences between how Kaoto behaves as a web deployment vs how it behaves as a VS Code target.

Updating `React` in the [`kie-tools`](https://github.com/apache/incubator-kie-tools) it's not viable at the moment, see: https://github.com/apache/incubator-kie-issues/issues/118

### Changes
This commit copies the `KogitoEditorEnvelope.tsx` and the `index.ts` file into this repository so we can override how React is bootstrapped to use the `New Root API`.

### Notes
* https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis
* https://github.com/reactwg/react-18/discussions/5

cc @fantonangeli

fix: https://github.com/KaotoIO/vscode-kaoto/issues/1074

### DIFF
--- a/package.json
+++ b/package.json
@@ -681,6 +681,7 @@
     "@types/fs-extra": "^11.0.4",
     "@types/mocha": "^10.0.10",
     "@types/react": "18.3.1",
+    "@types/react-dom": "18.3.1",
     "@types/vscode": "^1.100.0",
     "@types/wait-on": "^5.3.4",
     "@typescript-eslint/eslint-plugin": "^8.44.0",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@kie-tools-core/vscode-extension": "10.0.0",
     "@patternfly/patternfly": "6.3.1",
     "@patternfly/react-code-editor": "6.3.1",
-    "@patternfly/react-core": "6.3.1",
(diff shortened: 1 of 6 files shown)


apupier (changes requested):
tests related to custom catalogs are failing.

EDIT: Testing manually, at first sight the catalog are loading correctly. so might be just a UI test to adjust

djelinek:
in case you will need some help with failing UI tests, let me know 🙂

Issue #1095, opened 2025-09-24
Fix issue related to flaky test Deployments View -> Manipulate 'sample2.camel.yaml' routes -> click 'Resume'


apupier:
```
   1) Deployments View
       Manipulate 'sample2.camel.yaml' routes
         click 'Resume' on route-1151:
     TimeoutError: Timeout: route is not in an expected state - "Started"
Wait timed out after 10347ms
      at D:\a\vscode-kaoto\vscode-kaoto\node_modules\selenium-webdriver\lib\webdriver.js:929:22
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/613806ac-879a-4dff-86c1-e197b2c73c6f" />

apupier:
relatively similar issue occured:
```
 1) Deployments View
       Manipulate 'kaoto.camel.xml' routes
         click 'Stop' on route-1643:
     TimeoutError: Failed while waiting on terminal to has text: Stopped route-1643
Wait timed out after 10546ms
      at /Users/runner/work/vscode-kaoto/vscode-kaoto/node_modules/selenium-webdriver/lib/webdriver.js:929:22

  2) Deployments View
       Manipulate 'kaoto.camel.xml' routes
         click 'Start' on route-1643:
     TimeoutError: Timeout: route is not in an expected state - "Started"
Wait timed out after 10623ms
      at /Users/runner/work/vscode-kaoto/vscode-kaoto/node_modules/selenium-webdriver/lib/webdriver.js:929:22
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)

  3) Deployments View
       Manipulate 'kaoto.camel.xml' routes
         click 'Suspend' on route-1643:
     TimeoutError: Failed while waiting on terminal to has text: Suspended route-1643
Wait timed out after 12202ms
      at /Users/runner/work/vscode-kaoto/vscode-kaoto/node_modules/selenium-webdriver/lib/webdriver.js:929:22

  4) Deployments View
       Manipulate 'kaoto.camel.xml' routes
         click 'Resume' on route-1643:
     TimeoutError: Timeout: route is not in an expected state - "Started"
Wait timed out after 10452ms
      at /Users/runner/work/vscode-kaoto/vscode-kaoto/node_modules/selenium-webdriver/lib/webdriver.js:929:22
[...]

PR #1098, opened 2025-09-25
1095 add better error message

apupier:
relates to #1095


apupier:
need another approach, it wasn't able to retrieve more information in case of failure

PR #1196, opened 2025-12-10
feat(deploy): enable deployment to local kubernetes clusters

terryvel:
This PR adds support for deploying Camel integrations to a local Kubernetes cluster, in addition to the existing OpenShift deployment.

A new option is now available when triggering the "Deploy" command, allowing users to select the desired target platform:

- OpenShift (default behavior)
- Kubernetes (new)

If Kubernetes deployment is selected (either through a prompt or via VS Code Settings), the extension now generates a ShellExecution to run:

kamel run <filePattern>

This enables Kubernetes users  (especially those working with Minikube, Kind, K3d, or other local clusters) to deploy and test integrations directly from the Kaoto VS Code extension.

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -1,0 +1,118 @@
+# package.json
+
+diff --git a/package.json b/package.json
+index 972b6de..7a20baf 100644
+--- a/package.json
++++ b/package.json
+@@ -793,6 +793,15 @@
+             "tags": [
+               "preview"
+             ]
++          },
++          "kaoto.deployment.clusterType": {
++            "type": "string",
++            "markdownDescription": "Allow to select the cluster type for the deployment. If not set, Kaoto will ask via Quick Pick.",
++            "enum": [
++              "OpenShift",
++              "Kubernetes"
++            ],
++            "scope": "window"
+           }
+         }
+       },
+
+# src/extension/ExtensionContextHandler.ts
+
+	public registerKubernetesRunCommands() {
+		const INTEGRATIONS_KUBERNETES_RUN_COMMAND_ID: string = 'kaoto.integrations.kubernetes.run';
+
+		this.context.subscriptions.push(
+			vscode.commands.registerCommand(INTEGRATIONS_KUBERNETES_RUN_COMMAND_ID, async (integration: Integration) => {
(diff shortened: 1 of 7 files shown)


djelinek:
Hello! First of all, thank you for taking the time to make this contribution! 🙏🚀 

I actually have the same observation as Ricardo mentioned. 

1. Using the Camel K CLI (`kamel`), instead of the Camel JBang Kubernetes plugin.
Here, I believe we can migrate to using:
`camel kubernetes run <file> --cluster-type=kind/minikube/openshift`
See: https://camel.apache.org/manual/camel-jbang-kubernetes.html#_kubernetes_export

`--cluster-type      The target cluster type. Special configurations may be applied to different cluster types, such as Kind or Minikube.`

2. Using prompt selection as the default user setting for next time — In the first iteration, we can allow the user to select each time or predefine using his global/workspace extension settings. In an upcoming iteration, it could be beneficial to present a cache mechanism for these kinds of prompt steps. This could be handled using workspace context states, for example.

Overall, I like the idea. You've actually nailed it because it was on our to-do list as well. There will be an opportunity to implement a similar prompt step in the "Export to Maven Project" feature.

Thank you once again, and if there's anything I can help with, just let me know.

lordrip:
Hi @terryvel, we're happy it worked for you :ok_hand: and even more happy you took the time to put this PR together.

In the team, we were thinking that even if we prefer not moving forward with the `kamel` CLI bit, there's value in the remaining approach of asking the user what type of container orchestration service to use.

Here's a rough idea about what I mean:
1. Create a new setting to determine the cluster type:
    a. `ask` (default)
    b. `OpenShift`
    c. `Kubernetes`
    d. the remaining values

2. Since the default is `ask`, when the user triggers the deployment, we provide a list of the available container platforms.

3. We append that selection to the argument list

We won't need to save the setting automatically, since it's exposed through the VS Code settings page.

What do you think about this?

lordrip:
Hi @terryvel, thanks once again for your contribution. I'm writing to you to let you know that in the near future we're gonna migrate the extension build to Kaoto itself, and once we land there, we might revisit this ticket. After the move, please feel free to check the ticket out again if you're still interested.

Thanks once again for your contribution.

Issue #1468, opened 2026-08-10
IT: Kaoto editor webview renders blank on VS Code >= 1.123 (webview resources corrupted by startup race, then cached)


lordrip:
> Fully rewritten 2026-08-11 after the root cause was isolated. Part of #1466.

## What happens

Since VS Code 1.123.0, integration tests that open the Kaoto editor fail with `Kaoto topology was not loaded properly`: the editor webview renders blank. On CI this hits every OS with `CODE_VERSION: max`; locally it reproduces 100% against a long-lived `test-resources` dir.

The webview is blank because its JS bundle **arrives corrupted**: every 256 KiB chunk of the response is delivered twice (`chunk0 chunk0 chunk1 chunk1 ...`), so the 10.4 MB `KaotoEditorEnvelopeApp.js` arrives at ~20.9 MB, the doubled chunk splits a `const` from its initializer, the script throws `SyntaxError: Missing initializer in const declaration`, and React never mounts. All 23 resources the webview loads are affected (JSON catalogs, fonts, a 1 KB file doubles too). This is a transport bug — not a Kaoto rendering bug, not CSP, not the VSIX.

## Root cause

Two stacked problems, both upstream:

1. **VS Code startup race (microsoft/vscode#330243).** ExTester's `--open_resource` opens the test workspace via a second-instance CLI call (`cli.js -r <folder>`) right after launching VS Code through chromedriver. When that open lands while VS Code is still starting up, VS Code ≥ 1.123.0 registers a duplicate webview-resource reply channel (main log: `RequestStore#acceptReply was called without receiving a matching request N`, three times per request id) and every webview resource chunk in that window is delivered twice for the window's lifetime. It is a race: fast boots escape, slow boots corrupt — which is why CI looked platform-dependent and intermittent. VS Code 1.122.0 (Chromium 142) is immune.

2. **Service-worker cache poisoning makes it sticky.** The webview service worker caches the doubled response in the profile's `WebStorage/<partition>/CacheStorage`, and VS Code persists the webview origin per (viewType, extension) in `state.vscdb`, so every later session reuses the poisoned cache **without the race**. One bad boot permanently corrupts a `test-resources/settings` dir — that is why CI retry attempts fail deterministically after a raced first attempt, and why local runs against a months-old storage dir were 100% reproducible.

[...]

Issue #1466, opened 2026-08-10
Integration tests: CI failure triage grouped by root cause (from #1464 ExTester 8.24.0 bump)


lordrip:
## Context

The `integration-tests` CI jobs have been failing on [#1464](https://github.com/KaotoIO/vscode-kaoto/pull/1464) (bump `vscode-extension-tester` 8.23.0 → 8.24.0). This issue records the results of a full investigation of the failing runs so the distinct problems can be worked on independently.

**Scope note:** the original write-up below was diagnosis-only, deliberately proposing no fixes — the point was to separate the failures by root cause, because the raw failure counts conflate several unrelated problems with a large amount of cascade noise. Fixes are now landing on `fix/extester-update`; see **Progress** for current state and for the corrections that investigation produced.

Runs analysed (branch `chore/extester8.24.0`, 2026-08-08):

| Workflow | Stage | Result | Job |
|---|---|---|---|
| Build with **main** branch of Kaoto | `integration-tests / test-ubuntu` | FAILURE | `93120430393` |
| Build with **main** branch of Kaoto | `test-macos`, `test-windows` | skipped (`run-macos: false`, `run-windows: false`) | — |
| Build with **released** Kaoto | `integration-tests / test-ubuntu` | FAILURE | `93121287612` |
| Build with **released** Kaoto | `integration-tests / test-macos` | FAILURE | `93121287618` |
| Build with **released** Kaoto | `integration-tests / test-windows` | FAILURE | `93121287628` |

All `build`, SonarCloud and Snyk stages passed. Only integration tests fail.

Every stage runs under `nick-fields/retry@v4.0.0` with `max_attempts: 2`, `timeout_minutes: 30`, so each log contains two test runs. The attempts differ substantially:

| Stage | Attempt 1 | Attempt 2 |
|---|---|---|
| main-Kaoto / ubuntu | 77 passing / 22 failing, then **killed**: `Timeout of 1800000ms hit` | 14 passing / 42 failing |
| released / ubuntu | **78 passing / 21 failing** (28m) | 0 passing / 45 failing (5m) |
| released / macOS | 16 passing / 40 failing | 16 passing / 40 failing (identical) |
| released / windows | 2 passing / 40 failing (20m) | 2 passing / 40 failing (identical) |

The Ubuntu attempt-1 runs are the most informative; the others are heavily contaminated by the session-death cascade described in #1470.

---

## Work items

[...]

PR #1475, opened 2026-08-11
TEMP: isolate #1468 — CODE_VERSION A/B on webview suites (not for merge)

lordrip:
**Do not merge.** Diagnostic branch for #1468 (part of #1466).

This branch sits on the `vscode-extension-tester` 8.24.0 bump (tree-identical to `6d5e91f`) and trims CI so the blank-webview failure can be measured in isolation, away from the stacked fixes on `fix/extester-update` (#1465):

- Only the webview-dependent suites run: `BasicFlow`, `SwitchBetweenTabs`, `PropertyPanelLoading`, `ContextualMenuOpen`.
- `CODE_VERSION` is A/B'd in a matrix: `max` (VS Code ≥ 1.123.0, Chromium 148) vs `1.122.0` (Chromium 142) — 3 ubuntu samples + 1 windows sample per version.
- Single attempt per job (retry wrapper removed) so each job is one clean sample; minikube, macOS, unit tests and SBOM are dropped from the loop.

Expected outcome based on local measurements: all `1.122.0` jobs green; `max` jobs failing with `Kaoto topology was not loaded properly` because the webview bundle arrives with every 256 KiB chunk duplicated (see #1468 for the full analysis). This PR exists to measure the real CI failure rate on Chromium 148, since local reproduction is 100% deterministic but CI has shown intermittency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AtvAiEuFSkrYXehmdRdbHo

### DIFF
--- a/.github/workflows/_integration-tests.yaml
+++ b/.github/workflows/_integration-tests.yaml
@@ -1,3 +1,8 @@
+# TEMP (#1468): this branch trims CI to isolate the blank-webview reproduction.
+# Only the webview-dependent suites run (see the test:it:with-prebuilt-vsix glob),
+# each matrix job is a single attempt (no retry) so every job is one clean sample,
+# and CODE_VERSION is A/B'd between `max` (VS Code >= 1.123.0, Chromium 148) and
+# 1.122.0 (Chromium 142, known good). Not for merge.
 on:
   workflow_call:
     inputs:
@@ -68,9 +73,15 @@ jobs:
     if: ${{ inputs.run-ubuntu }}
     runs-on: ${{ inputs.runner-ubuntu }}
     timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        code-version: [max, "1.122.0"]
+        sample: [1, 2, 3]
+    name: ubuntu / ${{ matrix.code-version }} / sample ${{ matrix.sample }}
 
     env:
-      CODE_VERSION: max
+      CODE_VERSION: ${{ matrix.code-version }}
       TEST_RESOURCES: test-resources
 
     steps:
@@ -78,12 +89,6 @@ jobs:
       - uses: ./.github/actions/setup-tools
         with:
           cache: yarn
-      - uses: ./.github/actions/setup-minikube
-        with:
-          enable-qemu: ${{ inputs.enable-qemu && 'true' || 'false' }}
-      - name: Install Yarn
-        if: ${{ inputs.install-yarn }}
-        run: npm install -g yarn
       - name: Download VSIX artifact
         uses: actions/download-artifact@v8
         with:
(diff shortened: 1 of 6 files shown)


lordrip:
CI A/B results from the first two runs (webview suites only, single attempt per job):

| CODE_VERSION | Result |
| --- | --- |
| `1.122.0` (Chromium 142) | **8/8 pass** (6× ubuntu, 2× windows) |
| `max` = 1.132.0 (Chromium 148) | **5/7 fail** (4× ubuntu, 1× windows; 2 ubuntu samples passed) |

The intermittency on `max` is expected: the root cause has been isolated to a startup race, reproduced locally in a self-contained script with **no Kaoto, no kie-tools, no ExTester and no chromedriver**:

1. Launch VS Code; ~1 s later (while it is still booting) open a folder in the running instance via the CLI (`cli.js -r <folder>` — exactly what ExTester's `--open_resource` does after chromedriver launches VS Code).
2. On VS Code ≥ 1.123 this registers a duplicate webview-resource reply channel (`RequestStore#acceptReply was called without receiving a matching request` ×3 per request id in the main log). Every 256 KiB chunk of every webview resource is delivered twice → the Kaoto bundle arrives at ~2× its size and fails to parse → blank editor.
3. The webview service worker then **caches the doubled response**; the persisted webview origin (`memento/mainThreadCustomEditors.origins` in `state.vscdb`) routes every later session back to that cache, so a single bad boot permanently poisons the profile — which is why a retried CI job fails deterministically after a raced first attempt, and why long-lived local `test-resources/settings` dirs reproduce 100%.

Measured: 3/3 corrupt on 1.132.0 (ratios 2.0, 1.981, 1.981), 2/2 clean on 1.122.0 with the identical script. An upstream microsoft/vscode issue draft with the minimal repro is ready for review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lordrip:
Pushed `168df54`: a `yarn patch` for `vscode-extension-tester` that works around the root cause (microsoft/vscode#330243, redhat-developer/vscode-extension-tester#2454) by having `VSBrowser.openResources` wait for the workbench and settle for 10 s (once per session, `EXTESTER_OPEN_RESOURCE_SETTLE_MS` to override) before the first second-instance CLI open.

Verified locally on a fresh profile with VS Code 1.132.0: the ExTester repro corrupted 1/1 unpatched, clean 2/2 patched.

**This turns the matrix on this PR into the workaround's CI validation:** before the patch, `CODE_VERSION: max` failed 5/7 jobs while `1.122.0` passed 8/8. If the workaround holds, the `max` jobs should now go green too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lordrip:
**Workaround validated on CI.** With the ExTester patch (`168df54`), all 14 integration-test jobs are green across both workflow runs:

| CODE_VERSION | Before patch | After patch |
| --- | --- | --- |
| `max` (1.132.0, Chromium 148) | 5/7 failed | **7/7 pass** (6× ubuntu, 1× windows) |
| `1.122.0` (control) | 8/8 pass | 7/7 pass |

The `yarn patch` (wait for workbench + one-time 10 s settle before the first `--open_resource` CLI open) fully suppresses the microsoft/vscode#330243 race on real CI runners, with no version pin needed. Ready to be lifted into `fix/extester-update` / main, and to be proposed upstream as a fix for redhat-developer/vscode-extension-tester#2454.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1478, opened 2026-08-11
fix: replace extension recommendation UI test with a unit test

### LINKED ISSUE #1467: IT: notification lookups match the wrong notification, plus vscode-settings drift
Part of #1466 — see that epic for the CI runs analysed and the full triage context.

**Status: root cause confirmed and reproduced locally.**

## Failing tests
- `Extension > Check 'XML' extension recommendation exists`
- `Extension > Check 'YAML' extension recommendation exists`
- `Maven dependency update pom.xml > should invoke the update of the Maven dependencies when a step is added` (same defect, different call site)

## Stacktrace summary
```
NoSuchElementError: no such element: Unable to locate element:
  {"method":"xpath","selector":".//a[@role='button' and text()='Never']"}
  at async CenterNotification.takeAction (@redhat-developer/page-objects/out/components/workbench/Notification.js:119:9)
  at async Context.<anonymous> (out/default/extension.test.js:30:13)
```

## Problem

Six call sites in `it-tests/` look up a notification with an `async` predicate passed to `Array.prototype.find`:

```ts
const xml = notifications.find(async (n) => (await n.getText()).includes('XML Language Support by Red Hat'));
```

`find` does not await its predicate. An `async` callback always returns a `Promise`, every `Promise` is truthy, so `find` matches on the first iteration and returns `notifications[0]` regardless of content. The search string is never compared.

Sites: `it-tests/default/extension.test.ts` lines 46, 52, 60, 66, 75; `it-tests/MavenDependencyUpdate.test.ts` line 134.

Consequences:
- `expect(xml).to.not.be.undefined` can only fail when the notification centre is empty, so it passes in essentially all states.
- `takeAction('Never')` is invoked on whatever notification is first.
- The `try`/`catch` retry blocks re-run the identical broken lookup, so the retry can never succeed.
- `notificationAvailable()` has the same defect, reducing it to "is the array non-empty?".

## Contributing factor: settings drift

`it-tests/vscode-settings-minikube.json` has drifted from `it-tests/vscode-settings.json`. The minikube file — used only by the Ubuntu stages via `test:it:with-prebuilt-vsix:minikube` — is missing five keys, including `redhat.telemetry.enabled: false` and `git.openRepositoryInParentFolders: "never"`. Commits `3b1b36a` and `237e14b` each edited only the standard file.

The result is that two unrelated Info notifications (Red Hat telemetry, and the Git "repository found in parent folders" prompt) are present when the tests run, and the recommendation notifications are not yet.

Reproduced locally with a pristine VS Code profile, single variable:

| `--code_settings` | Notifications at ~5s | XML recommendation | Result |
|---|---|---|---|
| `vscode-settings.json` | XML + YAML | visible ~5s | 2 passing |
| `vscode-settings-minikube.json` | Git + telemetry (YAML at ~10s) | never in 120s | `NoSuchElementError` as in CI |

## Knock-on effect

These two tests are effectively cleanup — their comment is `// click 'Never' to avoid interruption with other tests`. Because they dismiss the wrong notification, the recommendation toasts survive. A screenshot from later in the same run shows three notification toasts stacked over the Kaoto canvas, one of them still the un-dismissed *"Kaoto recommends you install "YAML Language Support by Red Hat""*. This plausibly contributes to the `ElementClickInterceptedError` failures elsewhere, though that link is not yet proven.

Two further notification sources are visible in the same screenshot and are not dismissed by anything today:
- *"You are using Red Hat Camel catalog (4.18.1.redhat-00019) with JBang CLI. Specific Maven settings.xml configuration is required…"* (Source: Kaoto)
- *"Do you want to install the recommended 'Extension Pack for Java' extension from Microsoft for this repository?"*

lordrip:
Fixes https://github.com/KaotoIO/vscode-kaoto/issues/1467

## Problem

The recommendation toasts are shown once per session by RecommendationCore, on a 6s timer after activation, so no test can synchronise with them. The extension.test.ts only saw them by happening to run at the right moment. Its lookups also passed an async predicate to Array.prototype.find, which never awaits, so they always matched the first notification: they dismissed unrelated toasts and left the recommendations stacked over the canvas for later tests.

## Solution

- Suppress recommendation toasts suite-wide with `extensions.ignoreRecommendations`
- Close the drift in vscode-settings-minikube.json
- Cover `showRecommendedExtensions()` with a unit test instead
- Fix the same `find()` defect in MavenDependencyUpdate, where the test does trigger the notification it looks for

## Additional Changes

- First unit test to import ExtensionContextHandler, which type-checks it under strict for the first time
- Add definite-assignment assertions for properties used across methods
- Add `.bob-runs/` to .gitignore (AI agent tooling run logs)

Part of https://github.com/KaotoIO/vscode-kaoto/issues/1466

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of unsaved `pom.xml` warning notifications during Maven dependency updates.
  * Added coverage for XML and YAML extension recommendations.

* **Chores**
  * Updated development settings for a more consistent test environment.
  * Excluded AI delegation run logs from version control.
  * Removed obsolete extension integration tests and streamlined internal view management.


PVinaches (approved):
LGTM

PR #1485, opened 2026-08-12
fix: ensure Deployments tree retains route items and fix latent test defect

### LINKED ISSUE #1472: IT: Deployments tree loses its route items
Part of #1466 — see that epic for the CI runs analysed and the full triage context.

**Status: single occurrence, no error text captured.**

## Failing test
- `Deployments View > Manipulate 'kaoto.camel.xml' routes > click 'Resume' on route-1643` (main-ubuntu attempt 1 only)

The identical four-step sequence on the YAML route immediately above it passed in full (`Stop`, `Start`, `Suspend`, `Resume` on `route-1151`), and the same test **passes** in the released-Kaoto Ubuntu job (9.4s).

No stack trace is available: attempt 1 was killed by the 30-minute timeout before mocha printed failure details, and attempt 2 never reached this test.

The failure screenshot shows the Deployments view containing only `Localhost` with **nothing expanded beneath it** — no integration node and no route items — while the terminal shows `kaoto.camel.xml:12` logging once per second, i.e. the route is running normally. The route almost certainly did resume; the test had no tree item to act on.

## Separately: a latent defect in this test

The terminal assertion is:

```ts
waitUntilTerminalHasText(driver, [`${rm.state} ${p.route}`], …)
```

For the `Start` step that is `Started route-1643`; for the `Resume` step it is *also* `Started route-1643`. The terminal accumulates, so by the time `Resume` runs the string is already present from `Start` two steps earlier — that assertion cannot fail and is not testing the resume transition.

lordrip:
Fixes https://github.com/KaotoIO/vscode-kaoto/issues/1472

### Description

- Avoid releasing ports in `DeploymentsProvider` during background polling when transient fetch errors or timeouts occur (as they are already properly managed via task process lifecycles).
- Fix latent test defect in `09_DeploymentsViewRoutesManipulaton` by counting terminal occurrences of the target route status before and after executing route transitions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment route handling for tasks with valid network ports.
  * Added more reliable localhost route discovery by waiting for ports to become reachable.
  * Preserved allocated ports when route checks fail, preventing premature release.
  * Improved deployment terminal validation and resume behavior in automated checks.

### DIFF
--- a/it-tests/views/09_DeploymentsViewRoutesManipulaton.test.ts
+++ b/it-tests/views/09_DeploymentsViewRoutesManipulaton.test.ts
@@ -89,7 +89,7 @@ describe('Deployments View', function () {
 			routeManipulations.forEach((rm) => {
 				it(`click '${rm.button}' on ${p.route}`, async function () {
 					const textToLookFor = rm.terminalText !== null ? `${rm.terminalText} ${p.route}` : null;
-					const initialCount = textToLookFor !== null ? await getTerminalOccurrences(textToLookFor) : 0;
+					const initialCount = textToLookFor !== null ? await getInitialTerminalOccurrences(textToLookFor) : 0;
 
 					await clickRouteActionButton(rm.button);
 
@@ -102,6 +102,11 @@ describe('Deployments View', function () {
 			});
 		});
 
+		/**
+		 * Reads the current terminal occurrence count for `text`.
+		 * Used by the post-action polling loop — failures are treated as 0
+		 * so the poll keeps retrying without aborting the wait.
+		 */
 		async function getTerminalOccurrences(text: string): Promise<number> {
 			try {
 				const terminal = await activateTerminalView();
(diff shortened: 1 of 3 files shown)


lhein (dismissed):

lordrip:
I investigated the Windows failure in this PR and compared it with the recent released-version failure in PR #1487.

### Findings

This still looks like the same failure family tracked in #1466 under #1472 (`Deployments tree loses its route items`), but the current fix does not appear to fully address it on Windows.

What I found from the failing Windows run for this PR:

- only `integration-tests / test-windows` failed; build, Ubuntu, and macOS passed
- the failure is still localized to `Deployments View -> Manipulate 'sample2.camel.yaml' routes`
- failing assertions were:
  - `click 'Suspend' on route-1151`
  - `click 'Resume' on route-1151`
- both time out while trying to find/click the expected route action button in the tree

From the uploaded Kaoto logs, there is a strong repeated signal that the Deployments state is still going stale during route transitions:

- `[DeploymentsProvider] Timeout waiting for Camel integration on port -1`
- `Port -1 released and marked as free`
- stale old ports from prior runs are still being polled after shutdown (`Port 64613 not reachable`, `Port 64685 not reachable`)

I also checked PR #1487 (which touches unrelated startup logging) because it failed on Windows with a very similar shape. That run failed slightly later in the same route-manipulation flow: the `Resume` click happened, but the test then timed out waiting for route state `Started`. Its logs show the same invalid-port symptom:

- `[DeploymentsProvider] Timeout waiting for Camel integration on port -1`

### Interpretation

So this does not look like a regression caused by the code in #1487. Instead, both PRs point at the same underlying Deployments-state problem:

- stale deployment entries / route parent associations survive refreshes
- invalid or stale ports are still being tracked (`-1` and old stopped ports)
- on Windows that leaves the Deployments tree out of sync during route action transitions
- depending on timing, the visible symptom is either:
  - the expected action button never appears (`Suspend` / `Resume`), or
  - the click succeeds but the route state in the tree never reaches the expected label (`Started`)

### Suggested next steps

Before trying more blind test adjustments, I think the next step should be Windows-side debugging of the Deployments provider lifecycle itself:

1. trace where `-1` enters the deployment refresh / release path
2. inspect how route items keep their parent integration port across refreshes
3. verify whether stopped integrations from earlier test phases are still present in provider state when the route manipulation suite runs
4. add temporary logging around:
   - route action handlers
   - deployment tree rebuild / refresh
   - port release / reuse in `PortManager`
   - creation of `ParentItem` / `ChildItem`
5. reproduce on a Windows machine, since the timing-sensitive failure is only showing there in CI

I’m going to switch to a Windows environment next and debug it there, since this now looks more like a provider-state / lifecycle problem than a pure test issue.

lordrip:
I merged the PR to unblock the pipeline, I'll address the coderabbit comments in a new PR

